PR #28, opened 2019-01-29
 Change selector structure and add "astro" prefixes

larimaza:
# What

* Dissociate Astro from HTML structure and use CSS classes instead;
* Update Typography specs.

# Why

When trying to import Astro on legacy code projects, there were several breaks caused by our strict style rules, which caused us to decide to change implementation approach to something a little softer and more gradual/optional.

# How

* Change the CSS selection structure, from selecting HTML tags to using specific classes;
* Add `astro__` prefixes to classes in order to minimize conflicts with outer CSS styles;
* Update Typography to reflect recent design spec changes;
* Create `fonts.css` to store default primary and secondary font settings;
* Update Readme and make a few adjustments on the docs to reflect the changes.

### DIFF
--- a/docs/buttons.mdx
+++ b/docs/buttons.mdx
@@ -18,23 +18,23 @@ Use these for the main action in a section or screen. You should display only on
 Mouseover and click to view `:hover` and `:active` states.
 
 <Playground>
-  <button className="btn btn--uranus btn--medium">
+  <button className="astro__btn astro__btn--uranus astro__btn--medium">
     uranus
   </button>
 
-  <button className="btn btn--earth btn--medium">
+  <button className="astro__btn astro__btn--earth astro__btn--medium">
     earth
   </button>
 
-  <button className="btn btn--venus btn--medium">
+  <button className="astro__btn astro__btn--venus astro__btn--medium">
     venus
   </button>
 
-  <button className="btn btn--mars btn--medium">
+  <button className="astro__btn astro__btn--mars astro__btn--medium">
     mars
   </button>
   
-  <button className="btn btn--uranus btn--medium" disabled>
+  <button className="astro__btn astro__btn--uranus astro__btn--medium" disabled>
     disabled
   </button>
 </Playground>
@@ -48,23 +48,23 @@ Please note that there are color variations to the outline CSS classes. You also
 Mouseover and click to view `:hover` and `:active` states.
 
 <Playground>
-  <button className="btn btn--outline-uranus btn--medium">
+  <button className="astro__btn astro__btn--outline-uranus astro__btn--medium">
     outline uranus
   </button>
 
-  <button className="btn btn--outline-earth btn--medium">
+  <button className="astro__btn astro__btn--outline-earth astro__btn--medium">
     outline earth
   </button>
 
-  <button className="btn btn--outline-venus btn--medium">
+  <button className="astro__btn astro__btn--outline-venus astro__btn--medium">
     outline venus
   </button>
 
-  <button className="btn btn--outline-mars btn--medium">
+  <button className="astro__btn astro__btn--outline-mars astro__btn--medium">
     outline mars
   </button>
  
-  <button className="btn btn--outline-venus btn--medium" disabled>
+  <button className="astro__btn astro__btn--outline-venus astro__btn--medium" disabled>
     outline disabled
   </button>
 </Playground>
@@ -78,23 +78,23 @@ Aside from using the ghost button CSS class, you also need to specify size and c
 Mouseover and click to view `:hover` and `:active` states.
 
 <Playground>
-  <button className="btn btn--ghost-uranus btn--medium">
+  <button className="astro__btn astro__btn--ghost-uranus astro__btn--medium">
     ghost uranus
   </button>
 
-  <button className="btn btn--ghost-earth btn--medium">
+  <button className="astro__btn astro__btn--ghost-earth astro__btn--medium">
     ghost earth
   </button>
 
-  <button className="btn btn--ghost-venus btn--medium">
+  <button className="astro__btn astro__btn--ghost-venus astro__btn--medium">
     ghost venus
   </button>
 
-  <button className="btn btn--ghost-mars btn--medium">
+  <button className="astro__btn astro__btn--ghost-mars astro__btn--medium">
     ghost mars
   </button>
  
-  <button className="btn btn--ghost-venus btn--medium" disabled>
+  <button className="astro__btn astro__btn--ghost-venus astro__btn--medium" disabled>
     ghost disabled
   </button>
 </Playground>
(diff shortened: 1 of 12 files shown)


src/css/fonts.css
```diff
@@ -0,0 +1,5 @@
+:root {
+  --font-primary: 600 16px/1.5 Poppins, sans-serif;
```
kuroski: Hmm, if you want to use shorthand syntax, make sure you you all properties, you can check out more about that here: https://csswizardry.com/2016/12/css-shorthand-syntax-considered-an-anti-pattern/

In this case, you should provide: `font: font-style font-variant font-weight font-stretch font-size / line-heignt font-family`

It may be like:

`--font-primary: normal normal 600 normal 16px/1.5 Poppins, sans-serif`

docs/buttons.mdx
```diff
@@ -18,23 +18,23 @@ Use these for the main action in a section or screen. You should display only on
 Mouseover and click to view `:hover` and `:active` states.
 
 <Playground>
-  <button className="btn btn--uranus btn--medium">
+  <button className="astro__btn astro__btn--uranus astro__btn--medium">
```
kuroski: This `astro__` prefix may be confusing, especially if we are using BEM syntax.
Imagine the scenario where we are making something like a "card".

``` html
<div class="astro__card">
  <div class="astro__card__header"></div>
  <div class="astro__card__body"></div>
  <div class="astro__card__footer"></div>
</div>
```

This [breaks the rule](https://medium.com/fed-or-dead/battling-bem-5-common-problems-and-how-to-avoid-them-5bbd23dee319) that we must stick with only one level of elements.

And I think that it causes your component class chain to be massive, I could have:

``` html
<button className="astro__btn astro__btn--ghost-uranus astro__btn--medium astro__btn--outline-uranus">
</button>
```

And this is just using our classes, in my team we also need to include some tracking class or id or just want to change its spacing:

``` html
<button className="my-form-button astro__btn astro__btn--ghost-uranus astro__btn--medium astro__btn--outline-uranus js-tracking-button">
</button>
```

I think if you want to use a prefix, its nicer to use `astro-` so that we don't break the BEM element rule, or use a shorthand `a-` or `ast-`

``` html
<div class="astro-card">
  <div class="astro-card__header"></div>
  <div class="astro-card__body"></div>
  <div class="astro-card__footer"></div>
</div>
```

``` html
<button className="my-form-button a-btn a-btn--ghost-uranus a-btn--medium a-btn--outline-uranus js-tracking-button">
</button>

<button className="my-form-button ast-btn ast-btn--ghost-uranus ast-btn--medium ast-btn--outline-uranus js-tracking-button">
</button>
```

src/css/base.css
```diff
@@ -1,9 +1,6 @@
-body {
-  font-family: 'Poppins', sans-serif;
-  font-weight: 400;
-  font-size: 16px;
-  line-height: 1.5;
-  color: var(--color-moon-900);
+.astro__body { /* styles replicated in typography to allow isolation */
```
kuroski: Now that `body` was replaced with a class, it will be nice if we document it, just to the user knows that they must use this class.

docs/buttons.mdx
kuroski: :+1: for `a-`

src/css/typography.css
```diff
@@ -1,33 +1,66 @@
...
-  font-family: 'Lato', sans-serif;
+.astro__text--secondary--large {
+  font: var(--font-secondary);
   font-size: 24px;
+  color: var(--font-color);
 }
 
-p.text--secondary.text--small {
-  font-size: 16px;
+.astro__text--secondary--medium {
```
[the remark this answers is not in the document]
kuroski: Woah, I didn't notice that one.
Bem allow you to have `text--secondary text--medium` but not `text--secondary--medium`

src/css/fonts.css
larimaza: Whoa, TIL! Thanks for that article, shorthands were always unclear territory for me and now it all makes more sense.

src/css/base.css
larimaza: Well they mustn't... 😬 I repeated all of the properties contained in `.a-body` in each typography element, so that the body class usage is not mandatory, because that would force these styles on an entire page.  
Think I left this here thinking more about the future than the present use...

src/css/typography.css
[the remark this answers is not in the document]
larimaza: Umm I think we're not following BEM literally, just as a guideline... In this case we do need the two modifiers so that we don't have to create two prefixed classes and make usage unnecessarily complicated. Do you guys think it's okay to skip that rule this time?

src/css/fonts.css
larimaza: Done on 181fa82

docs/buttons.mdx
larimaza: Agreed! Done on 181fa82

src/css/typography.css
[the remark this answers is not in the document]
elmarques: I'm kind of person who thinks it's cool to be flexible. 😝

It's ok for me to skip this rule this time.

But if we decide to follow BEM conventions we don't need to create two modifiers. I think BEM also supports compound names in modifiers (separated by dashes).

So `a-text--secondary-medium` would be 100% BEM compliance.

Btw, in some cases, devs also don't follow BEM by the book:
https://webuild.envato.com/blog/chainable-bem-modifiers/

docs/buttons.mdx
```diff
@@ -162,10 +162,10 @@ Although button classes are designed for `<button>` elements, you may also use t
 If your `<a>` button triggers in-page functionalities instead of linking (to other pages or within the current page), make them semantically accessible by adding a `role="button"`.
 
 <Playground>
-  <a className="btn btn--uranus btn--medium">
+  <a className="a-btn a-btn--uranus a-btn--medium">
     link button
   </a>
-  <a className="btn btn--uranus btn--medium disabled">
+  <a className="a-btn a-btn--uranus a-btn--medium disabled">
```
elmarques: Do you think `.disabled` still a generic name?

Maybe we can use something more Astro-BEM like:
```suggestion
  <a className="a-btn a-btn--uranus a-btn--medium a-btn--disabled">
```

src/css/typography.css
```diff
@@ -1,33 +1,66 @@
...
+  font: var(--font-primary);
+  font-size: 12px;
+  color: var(--font-color);
 }
 
 /* Secondary font */
 
-p.text--secondary {
-  font-family: 'Lato', sans-serif;
+.a-text--secondary--large {
```
kuroski: Hmm this case you have two modifiers in one...
It's nice to rename that to `.a-text--secondary-large` or split in `.a-text--secondary` and `.a-text--large`

And use the rule:
``` css
.a-text--secondary.a-text--large {
  ...
}
```

larimaza: Yes, I thought about this after El's comment and now I agree that `secondary-large` can be a compound modifier with two aspects, not two separate modifiers. Did that, and also refactored typography.css with a new idea to make it less repetitive, in this commit: 23fa82d

src/css/typography.css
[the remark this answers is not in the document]
larimaza: I thought about this again by reading this new reply and now I understand that this can be a compound modifier with two aspects instead of two modifiers. Agreed and changed it to a single dash. Also refactored typography.css with a new idea to make it less repetitive, in this commit: 23fa82d

docs/buttons.mdx
larimaza: I had to think about that, but I agree that some other CSS stylesheet may have a `.disabled` class and that would NOT be pretty lol 80336a7

PR #33, opened 2019-02-08
Add toggle component to the Astro

elmarques:
This PR adds a toggle component to the Astro Design System with _on_, _off_, _hover_, and _disabled_ states.

Also updates Control & Toggles docs.

captura de tela 2019-02-08 as 10 18 51

### DIFF
--- a/src/css/radio.css
+++ b/src/css/radio.css
@@ -9,7 +9,7 @@
 
 .a-radio > label {
   display: flex;
-  font-family: var(--font-secondary);
+  font: var(--font-secondary);
   font-size: 16px;
   align-items: top;
 }
--- a/src/css/radio.css
+++ b/src/css/radio.css
@@ -10,7 +10,6 @@
 .a-radio > label {
   display: flex;
   font: var(--font-secondary);
-  font-size: 16px;
   align-items: top;
 }
 
@@ -58,4 +57,11 @@
 
 .a-radio > input[type='radio']:hover:not(:disabled) + label::before {
   border-color: var(--color-uranus-500);
+}
+
+/* Cursor */
+
+.a-radio > input[type='radio']:hover:not(:disabled),
+.a-radio > input[type='radio']:hover:not(:disabled) + label {
+  cursor: pointer;
 }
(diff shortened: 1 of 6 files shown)


elmarques:
[💻: 1, Criar componentes de "toggles" e adicioná-los em "Controls & Toggles" na documentação](https://trello.com/c/9kHmaQyQ/355-%F0%9F%92%BB-1-criar-componentes-de-toggles-e-adicion%C3%A1-los-em-controls-toggles-na-documenta%C3%A7%C3%A3o)

src/css/toggle.css
```diff
@@ -0,0 +1,87 @@
...
+  height: var(--toggle-base-height);
+  vertical-align: middle;
+}
+
+/* Label */
+
+.a-toggle > label {
+  margin-right: 8px;
+  font-family: var(--font-secondary);
+  font-size: 16px;
```
brunobc182: I think in this case you can use the shorthand: `font: var (--font-secondary)`, because `--font-secondary` already have `font-size` on this same value.

elmarques: Great 👀, thanks! Fixed in ef304fe commit.

src/css/radio.css
```diff
@@ -9,7 +9,7 @@
 
 .a-radio > label {
   display: flex;
-  font-family: var(--font-secondary);
+  font: var(--font-secondary);
   font-size: 16px;
```
larimaza: Since you're touching this file.... we can remove this line as well 😬 standard secondary font size is 16px

src/css/toggle.css
```diff
@@ -0,0 +1,88 @@
...
+.a-toggle > input[type='checkbox']::after {
+  position: absolute;
+  top: 1px;
+  left: 1px;
+  display: block;
+  visibility: visible;
+  width: var(--toggle-circle-size);
+  height: var(--toggle-circle-size);
+  content: '';
+  transition: left 100ms ease-in-out;
```
larimaza:
```suggestion
  transition: left 0.3s ease;
```

PR #35, opened 2019-02-15
Use yarn

luisfmsouza:
# What

Start using yarn to unify our dependency management.

# Why

This is needed to avoid mismatch of package versions, specially when we add some package using `npm`, and someone else install dependencies using `yarn`. It'd cause us trouble!

# Changes

- Remove all code & files related to NPM;
- Upgrade package versions to the latest;
- Deduplicate `yarn.lock` file.


larimaza:
Yeeeeah I don't have a lot of context to review this haha but should we update the documentation in this same PR? It mentions NPM in a few places.

luisfmsouza:
@larimaza for sure! I've updated a few places, can you point me where I still need to update?

larimaza:
All good! 🚢

PR #37, opened 2019-02-19
Create standard inputs

brunobc182:
# What
Create the standard input component and add it to the documentation.

# How
- Create `inputs.css` with all the styles;
- Create `inputs.mdx` to add sliders to the docs;
- Create `js` folder and inputs.js` file to add JS to the component to manipulate floating labels and create a working sample in our docs.

# Sample
screenshot from 2019-02-19 16-57-21

# Staging
http://762abbf5.ngrok.io/#/

# Note
I'm still having problems with JS. That seems to load my listeners to control labels only when i reload the page. Do you guys have any thought about it?


Auralcat:
[📝: 1, 💻: 1, Criar componentes de "Standard Text Input" e página "Inputs" na documentação](https://trello.com/c/KAxT7R5Y/366-%F0%9F%93%9D-1-%F0%9F%92%BB-1-criar-componentes-de-standard-text-input-e-p%C3%A1gina-inputs-na-documenta%C3%A7%C3%A3o)

src/css/inputs.css
```diff
@@ -0,0 +1,43 @@
...
+  left: 16px;
+  transition: 0.2s ease all;
+  pointer-events: none;
+  color: var(--color-moon-300);
+  font: var(--font-secondary);
+}
+
+.a-input > input:focus {
+  border: 1px solid var(--color-uranus-500);
+  outline: none;
```
luisfmsouza: For accessibility purposes, we must never remove the outline

src/css/inputs.css
```diff
@@ -0,0 +1,43 @@
...
+}
+
+.a-input > input:focus + label {
+  top: 5px;
+  color: #159ce4;
```
luisfmsouza: Is this an Astro color? If not, can we use an Astro color here?

src/css/inputs.css
```diff
@@ -0,0 +1,43 @@
+.a-input {
+  position: relative;
+  margin-bottom: 46px;
+}
+
+.a-input > input {
+  width: 342px;
+  height: 56px;
```
luisfmsouza: I wondering about the width and height hard coded values. 

I think we can avoid the height declaration using the padding to compensate it. 

But about the width, looks like we can have something flexible to use even in mobile or another situations, maybe the container should be responsible for this? Wyt @larimaza ?

assets/icons/invalid-input.svg
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
luisfmsouza: I have 3 suggestions for this svg:

1) we should use a minifier tool to remove unnecessary code and also compress our SVG. We can add some tool to our build process or even use an online one like this: https://jakearchibald.github.io/svgomg/

2) the file name should be more related to the icon itself instead of the use of this icon, so my suggestion is to change to something like: check.svg (I'm using github icons as a reference: https://octicons.github.com/icon/check/)

3) the decoration of the SVG should be done via CSS: the green circle could be an div that wrap the check icon and the green color of the icon can be defined via CSS property `color: var(--some-green)` (I'm also following github's example here)

assets/icons/validated-input.svg
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
luisfmsouza: I have 3 suggestions for this svg:

1) (same of the last one) 

2) suggestion: i.svg

3) (same of the last one)

docs/inputs.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+  <div className="a-input a-input--validated">      
+    <input type="text" defaultValue="Astro Team" />
+    <label>Validated</label>
+  </div>
+
+  <div className="a-input a-input--invalid">      
+    <input type="text" defaultValue="Astro Team"/>
+    <label>Invalid</label>
+    <span className="a-input__error">Esse dado é inválido</span>
```
luisfmsouza: Just for cohesion, we should use an english message here: Invalid data

src/css/inputs.css
```diff
@@ -0,0 +1,119 @@
...
+  line-height: 15px;
+}
+
+.a-input--validated > input:focus + label {
+  color: var(--color-earth-400);
+}
+
+.a-input--validated > input:focus {
+  border: 1px solid var(--color-earth-400);
+  outline: none;
```
luisfmsouza: For accessibility purposes, we must never remove the outline

src/css/inputs.css
```diff
@@ -0,0 +1,119 @@
...
+
+.a-input > input:focus + label {
+  top: 5px;
+  color: var(--color-uranus-500);
+  font-size: 12px;
+  font-weight: bold;
+  line-height: 15px;
+}
+
+.a-input > .a-input--floating-label {
```
luisfmsouza: Why do we need a new class to move the label upward? Doesn't the `:focus` fit this need?

assets/icons/invalid-input.svg
kuroski: It's a very good suggestion.
You guys could see if we can use the `fill` property of the css
So then you can put the default color as green
And able users to create an css selector and change the color by need:
https://css-tricks.com/cascading-svg-fill-color/

docs/inputs.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+## standard text inputs
+
+The regular input. Labels appear on the default state and slightly move upwards when the user starts interacting and typing on the component.
+
+The floating label will need JS to work on your project. You may find the code for this working example in our GitHub repository.
+
+<Playground>
+  <div className="a-input">      
+    <input type="text"/>
```
kuroski: I think it's nice to put in the doc that the input field order IS important.
If the user put the `label` before the `input` that may be cause unwanted styling

src/css/inputs.css
kuroski: :+1: to that
element must be flexible

the responsibility of the input positioning must be from who will use this component

src/css/inputs.css
```diff
@@ -0,0 +1,119 @@
...
+}
+
+.a-input--validated::after {
+  position: absolute;
+  top: 12px;
+  right: 16px;
+  width: 32px;
+  height: 32px;
+  content: '';
+  background: url('../../assests/icons/validated-input.svg') center no-repeat;
```
kuroski: That wont cause any problems to who will use this input when they try to load that with this hard coded background?

And try to avoid shorthands if you are not using all properties, this is an anti pattern, especially with the background css property

You can check out more here: https://csswizardry.com/2016/12/css-shorthand-syntax-considered-an-anti-pattern/

src/css/inputs.css
```diff
@@ -0,0 +1,119 @@
...
+}
+
+.a-input--invalid > input:focus + label {
+  color: var(--color-mars-500);
+}
+
+.a-input--invalid > input:focus {
+  border: 1px solid var(--color-mars-500);
+  outline: none;
```
kuroski: try to avoid removing outline for accessibility purposes

if you want to give outline only when keyboard navigating you can see that behaviour here:
- https://css-tricks.com/keyboard-only-focus-styles/
- https://stackoverflow.com/questions/31402576/enable-focus-only-on-keyboard-use-or-tab-press#answer-45191208

We don't have a wide support, but it's a start

src/css/inputs.css
```diff
@@ -0,0 +1,119 @@
...
+}
+
+.a-input > .a-input__error {
+  position: absolute;
+  display: block;
+  margin: 4px 16px;
+  color: var(--color-mars-500);
+  font: var(--font-secondary);
+  font-size: 12px;
+  line-height: 18px;
```
kuroski: why you are setting the `line-height` here?
I remember that was a rule that says:
> all texts must be `1.5x` the font sizes

And we already set the overall page line-height to `1.5`
This affirmation is right?
summon @larimaza

larimaza: Yes, that's right. We shouldn't change line-height, if we need to adjust the font vertical space it's better to use padding!

src/css/inputs.css
larimaza: We already have a card in the Astro board to address that outline focus solution for the whole project, so we don't need to do it in this PR. We can just remove that `outline: none` line :)

src/css/inputs.css
larimaza: I tried to build this branch and there was an error with this path, maybe there's a typo in there? "assests" 🤔

larimaza: As for the background image, we could go back to this file and improve it once we implement all icons, I can put a note for that in the Icons card on Trello.

src/css/inputs.css
brunobc182: Done on 2352d24bc8eba943fa65aadf9a8856e86c82c477

src/css/inputs.css
brunobc182: Done on 2352d24bc8eba943fa65aadf9a8856e86c82c477

src/css/inputs.css
brunobc182: Done on 2352d24bc8eba943fa65aadf9a8856e86c82c477

assets/icons/invalid-input.svg
brunobc182: Done on 2352d24bc8eba943fa65aadf9a8856e86c82c477

We have done this using `mask-image`, because the attribute `fill` does not working with `background-image`. This solution came with a problem, we do not have IE support. Check on: https://caniuse.com/#search=mask-image

What you guys think?

assets/icons/validated-input.svg
brunobc182: Done on 2352d24bc8eba943fa65aadf9a8856e86c82c477

src/css/inputs.css
brunobc182: Done on 2352d24bc8eba943fa65aadf9a8856e86c82c477

src/css/inputs.css
brunobc182: We created this class for manipulate the floating label. Using only `:focus` we can't stick the label at the top when user don't have focus on it. So, do you have a better solution?

src/css/inputs.css
brunobc182: Done on 2352d24bc8eba943fa65aadf9a8856e86c82c477

src/css/inputs.css
brunobc182: Done on 2352d24bc8eba943fa65aadf9a8856e86c82c477

docs/inputs.mdx
brunobc182: Done on 2f9aee4568f3f0181437a0815f7ab3d6f81270ee

src/css/inputs.css
brunobc182: Done on 2352d24bc8eba943fa65aadf9a8856e86c82c477

docs/inputs.mdx
larimaza: > Awesome! Is there any staging we can see the inputs working?

I think it's easier to enter this branch and run it on your machine

docs/inputs.mdx
```diff
@@ -0,0 +1,43 @@
...
+import '../src/index.css';
+import { Playground } from 'docz';
+import './js/inputs';
+
+# Inputs
+
+## standard text inputs
+
+The regular input. Labels appear on the default state and slightly move upwards when the user starts interacting and typing on the component.
+
```
larimaza: Following Kuroski's comment, here's a suggestion:

```suggestion

Please follow the markup structure below in the same order to get all correct CSS styles.

```

src/css/inputs.css
larimaza: I think Bruno's right, once you move the focus out of the input the label would go back to the middle.

src/css/inputs.css
```diff
@@ -0,0 +1,114 @@
...
+  border-radius: 8px;
+  background-color: var(--color-space-100);
+}
+
+.a-input > label {
+  position: absolute;
+  top: 16px;
+  left: 16px;
+  transition: 0.2s ease all;
+  pointer-events: none;
```
larimaza: Just curious, what does `pointer-events` do here? I didn't know this lol

kuroski:
Just one thing
problem
I found out one small problem.
When the input have svg icon, maybe you should put some right padding to prevent the text to get bellow the icon

src/css/inputs.css
```diff
@@ -0,0 +1,114 @@
...
+
+.a-input--validated::after {
+  position: absolute;
+  top: 12px;
+  right: 16px;
+  width: 32px;
+  height: 32px;
+  content: '';
+  mask-image: url('../../assets/icons/check.svg');
+  background-color: var(--color-earth-500);
```
larimaza: I just checked the specs because the shades of green looked a little off, and looks like this should be earth-400 😬

src/css/inputs.css
luisfmsouza: Yes, you are right!

brunobc182:
> there

brunobc182:
> Just one thing
> problem
> I found out one small problem.
> When the input have svg icon, maybe you should put some right padding to prevent the text to get bellow the icon

Done on a78eac4bd9861802202dc5b5892bd071ba69971b

src/css/inputs.css
brunobc182: Precisamos disso para evitar que os usuários cliquem no rótulo. Sem isso, eles não podem selecionar corretamente a entrada.

assets/icons/invalid-input.svg
larimaza: I think we can leave it like this for now and merge this card, as we're still developing an official implementation for SVG's and we can come back to refactor it.

PR #38, opened 2019-02-20
Fix disabled state inconsistencies in Controls & Toggles

larimaza:
# What
Fix a few disabled state color inconsistencies in Controls & Toggles components.

# Why
In those components, the shades of grey in their disabled states were a little inconsistent with the design specs.

# How
* Remove all use of `opacity` in disabled states. Change the component colors directly instead, to `moon 200`, which is the official disabled state color.
* Remove the blue color in the `on disabled` toggle component, after checking with design, because from their UX point of view it's best to keep all disabled stuff in grey.
* Also cover the scenarios in which the same component could be `checked` and `disabled`, which was not covered yet.

They will all look exactly the same now <3

**Before** | **After**
-|-
screen shot 2019-02-20 at 16 01 33|screen shot 2019-02-20 at 16 01 10
screen shot 2019-02-20 at 16 01 28|screen shot 2019-02-20 at 16 01 16
screen shot 2019-02-20 at 16 01 48|screen shot 2019-02-20 at 19 13 22

### DIFF
--- a/docs/controls-toggles.mdx
+++ b/docs/controls-toggles.mdx
@@ -96,13 +96,8 @@ Mouseover to view `:hover` state.
   </div>
 
   <div className="a-toggle a-toggle--disabled">
-    <label htmlFor="toggle3">off disabled</label>
+    <label htmlFor="toggle3">disabled</label>
     <input type="checkbox" id="toggle3" disabled />
-  </div>
-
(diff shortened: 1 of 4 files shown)


larimaza:
https://trello.com/c/48yItRAn/102-checar-cor-das-labels-dos-elementos-em-controls-toggles

PR #39, opened 2019-02-22
Create a component to handle floating label and slider value 

brunobc182:
# What
Fix floating label on inputs page and slider value on controls & toggles page.

# Why
The floating labels don't work properly when we navigate at docs. They only work when we reload the entire page. And slider value doesn't work the way it should changed it.

# How
- Create a component on `input.js` to call at `input.mdx` file.
- Create a component on `slider.js` to call at `controls-toggle.mdx` file.

### DIFF
--- a/docs/inputs.mdx
+++ b/docs/inputs.mdx
@@ -2,9 +2,10 @@
 name: Inputs
 ---
 
+import {Fragment} from 'react'
 import '../src/index.css';
 import { Playground } from 'docz';
-import './js/inputs';
+import HandleFloatingLabel from './js/inputs';
 
 # Inputs
 
@@ -15,6 +16,7 @@ The regular input. Labels appear on the default state and slightly move upwards 
 Please follow the markup structure below in the same order to get all correct CSS styles.
 
 The floating label will need JS to work on your project. You may find the code for this working example in our GitHub repository.
+
 
 <Playground>
   <div className="a-input">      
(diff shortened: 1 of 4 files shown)


brunobc182:
[Investigar problema de carregamento de JS na documentação](https://trello.com/c/lB9WtW96/380-investigar-problema-de-carregamento-de-js-na-documenta%C3%A7%C3%A3o)

docs/inputs.mdx
```diff
@@ -2,9 +2,10 @@
 name: Inputs
 ---
 
+import {Fragment} from 'react'
```
larimaza: Just 💅 hahaha
```suggestion
import { Fragment } from 'react'
```

docs/js/slider.js
```diff
@@ -1,12 +1,20 @@
...
       const sliderValue = slider.value;
 
       slider.style.setProperty("--slider-percentage", `${sliderValue}%`);
       counter.innerText = `${sliderValue} coins`;
     });
-  })
+  });
+};
+
+const HandleSliderValue = () => {
```
kuroski: Maybe you could change to do that

``` javascript
const HandleSliderValue = changeSliderValue;
```

Or just

``` javascript
export default changeSliderValue
```

brunobc182: I did this because someone could just import the function and we need the component way to put on `mdx` file.

PR #43, opened 2019-02-25
Extract content from docs homepage to React components

larimaza:
# What
Prepare docs homepage structure for when we setup our own Docz theme.

# Why
The current structure mixing HTML and Markdown is really fragile and buggy - sometimes it parses MD correctly, sometimes it doesn't. That approach also doesn't allow a flex container in the "Get Started" section, because there's a code snippet in Markdown and it's not supported if it's inside HTML.

# How
* Extract content to React components when HTML is necessary.
* Add a new dependency (that will not go into the final bundle) called `babel-plugin-inline-import` that makes it possible to import a `.text` file into a React component and render pure text, instead of rendering `<button>` like an actual button. That was the only way to display a code snippet without Markdown 😬 
* Note: `home.css` is temporary. That styling will go into our theme later.

### DIFF
--- a/docs/components/GithubButtons.jsx
+++ b/docs/components/GithubButtons.jsx
@@ -1,0 +1,21 @@
+import React, { Fragment } from "react";
+
+export default () => (
+  <Fragment>
+    <iframe
+      src="https://ghbtns.com/github-btn.html?user=magnetis&repo=astro&type=star&count=true&size=large"
+      frameBorder="0"
+      scrolling="0"
+      width="160px"
+      height="30px"
+    />
+
+    <iframe
+      src="https://ghbtns.com/github-btn.html?user=magnetis&repo=astro&type=fork&count=true&size=large"
+      frameBorder="0"
+      scrolling="0"
+      width="160px"
+      height="30px"
+    />
+  </Fragment>
+);
--- a/docs/components/IntroCards.jsx
+++ b/docs/components/IntroCards.jsx
@@ -8,7 +8,7 @@ export default () => (
       <p>
         Astro is built based on <b>Atomic Design</b>, so our designers and
         developers can use its style set and components to enhance our product
-        interface, as a <b>modular system</b> that can be easily   <b>managed and updated</b>.
+        interface, as a <b>modular system</b> that can be easily <b>managed and updated</b>.
       </p>
     </div>
 
--- a/docs/components/IntroCards.jsx
+++ b/docs/components/IntroCards.jsx
@@ -16,7 +16,7 @@ export default () => (
       <h3>open</h3>
 
       <p>
-        Magnetis believes that good design and experiences can   <b>impact the product market and society itself</b>. Our <b>design and code assets are open</b> so we can contribute to the
+        Magnetis believes that good design and experiences can <b>impact the product market and society itself</b>. Our <b>design and code assets are open</b> so we can contribute to the
         community and help new product teams.
       </p>
     </div>
--- a/docs/components/IntroCards.jsx
+++ b/docs/components/IntroCards.jsx
@@ -1,0 +1,36 @@
+import React from "react";
+
+export default () => (
+  <div className="cards-box">
+    <div>
+      <h3>atomic</h3>
+
+      <p>
+        Astro is built based on <b>Atomic Design</b>, so our designers and
+        developers can use its style set and components to enhance our product
+        interface, as a <b>modular system</b> that can be easily   <b>managed and updated</b>.
+      </p>
+    </div>
+
+    <div>
+      <h3>open</h3>
+
+      <p>
+        Magnetis believes that good design and experiences can   <b>impact the product market and society itself</b>. Our <b>design and code assets are open</b> so we can contribute to the
+        community and help new product teams.
+      </p>
+    </div>
+
+    <div>
+      <h3>accessible</h3>
+
+      <p>
+        Design is meant to be <b>for all</b>. Astro style and components follow
+        basic recommendations from the <b>WCAG guidelines</b> for accessibility,
+        with tested contrast ratios and good practices in legibility and screen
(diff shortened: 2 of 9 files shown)


larimaza:
p.s.: not sure why all these "integrity" lines were removed from `yarn.lock` 🤔

docs/components/GithubButtons.jsx
```diff
@@ -0,0 +1,21 @@
+import React, { Fragment } from "react";
+
+export default () => (
+  <Fragment>
+    <iframe
+      src="https://ghbtns.com/github-btn.html?user=magnetis&repo=astro&type=star&count=true&size=large"
+      frameBorder="0"
+      scrolling="0"
+      width="160px"
```
elmarques: Almost unnecessary FYI moment 😆: `px` units are optional here.

docs/components/IntroCards.jsx
```diff
@@ -0,0 +1,36 @@
+import React from "react";
```
elmarques: OCD moment 😆: Is `"`double`"` quote our default to JS files? The .mdx files have `'`single`'` ones .

larimaza: GOD I don't even know anymore, this is driving me crazy too :( I can never get the linter to work

docs/components/GithubButtons.jsx
larimaza: lol true

docs/components/IntroCards.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+export default () => (
+  <div className="cards-box">
+    <div>
+      <h3>atomic</h3>
+
+      <p>
+        Astro is built based on <b>Atomic Design</b>, so our designers and
+        developers can use its style set and components to enhance our product
+        interface, as a <b>modular system</b> that can be easily   <b>managed and updated</b>.
```
kuroski:
```suggestion
        interface, as a <b>modular system</b> that can be easily <b>managed and updated</b>.
```

docs/components/IntroCards.jsx
```diff
@@ -0,0 +1,36 @@
...
+      </p>
+    </div>
+
+    <div>
+      <h3>open</h3>
+
+      <p>
+        Magnetis believes that good design and experiences can   <b>impact the product market and society itself</b>. Our <b>design and code assets are open</b> so we can contribute to the
```
kuroski:
```suggestion
        Magnetis believes that good design and experiences can <b>impact the product market and society itself</b>. Our <b>design and code assets are open</b> so we can contribute to the
```

docs/components/StartedCards.jsx
```diff
@@ -0,0 +1,82 @@
...
+        <li>
+          <p>
+            Install via terminal:
+            <br />
+            <code>yarn add @magnetis/astro</code>
+          </p>
+        </li>
+
+        <li>
+          <p>
```
kuroski: Maybe something like

``` html 
<span>Get our font links in the</span> 
<a href="https://magnetis.github.io/astro/#/docs-typography">
  Typography page
</a> 
<span>and load them in your project.</span>
```

docs/css/home.css
```diff
@@ -1,7 +1,18 @@
...
+.cards-box {
   display: flex;
+  justify-content: space-between;
 }
 
-.intro-box > div:not(:last-child) {
-  padding-right: 30px;
-}
\ No newline at end of file
+.cards-box > div {
```
kuroski: Hmmm here maybe it's a nice use case to css grid.

``` css
.cards-box {
  display: grid;
  grid-template-columns; repeat(3, 1fr);
  grid-column-gap: 15px; // I dont know how much space is needed
}
```

And then you can remove this `.cards-box > div` selector

package.json
```diff
@@ -31,6 +31,7 @@
   },
   "devDependencies": {
     "autoprefixer": "^9.1.5",
+    "babel-plugin-inline-import": "^3.0.0",
```
kuroski: It's not a problem using plugins, just watch out for not overusing babel and its plugins.
We can talk about this point if you want in a call, but for now it's not a problem >_<

docs/components/IntroCards.jsx
elmarques: 😆 

Do you use [Prettier](https://github.com/prettier/prettier)?

I usually put a `.prettierrc` file into root folder with some configs:

```
{
  "singleQuote": true,
  "printWidth": 120
}
```

docs/components/IntroCards.jsx
larimaza: ab76075

docs/components/IntroCards.jsx
larimaza: 6c51f9d

docs/css/home.css
larimaza: This CSS will be modified very soon, can I keep these suggestions for the new theme settings? :)

package.json
larimaza: Yes I agree, I just didn't find any other option after a lot of research here :/

docs/components/StartedCards.jsx
larimaza: Why do you say that? 🤔

package.json
kuroski: I understand =D
But for now it's not a problem 
It's just a thing to keep an eye open in the future to prevent from adding a lot of plugins.

docs/components/IntroCards.jsx
larimaza: Yeah, we don't have a .prettierrc file. Let's tackle that in another card, shall we? :D

docs/components/StartedCards.jsx
larimaza: Why do you say that? 🤔

PR #51, opened 2019-03-08
Refactoring components to standardize disabled state and add accessibility properties.

brunobc182:
# What

Refactoring components to standardize disabled state and add accessibility porperties.

# Why

Accessibility purposes and standardize disabled states.

# How

- Change radio label order to above input
- Change checkbox label order to above input
- Change slider label order to above input and create `a-slider--disabled` class
- Add accessibility properties on input html component

### DIFF
--- a/docs/controls-toggles.mdx
+++ b/docs/controls-toggles.mdx
@@ -18,6 +18,8 @@ Please follow the markup structure from the examples below. You must create a un
 
 Mouseover to view `:hover` state.
 
+To disabled state, add the `a-radio--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
+
 <Playground>
   <div className="a-radio">
     <label htmlFor="radio1">Radio</label>
@@ -62,6 +64,8 @@ Checkboxes are used when a user might select multiple options from a list, or wh
 
 Mouseover to view `:hover` state.
 
+To disabled state, add the `a-checkbox--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
+
 <Playground>
   <div className="a-checkbox">
     <label htmlFor="checkbox1">Checkbox</label>
@@ -89,6 +93,8 @@ Mouseover to view `:hover` state.
 A simple control that is used to quickly switch between two possible states.
 
 Mouseover to view `:hover` state.
+
+To disabled state, add the `a-toggle--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
 
 <Playground>
   <div className="a-toggle">
@@ -118,6 +124,8 @@ We recommend you set the slider width in the `a-slider` class.
 The colored percentage fill and counter label will need JS to work on your project. You may find the code for this working example in our GitHub repository. You can also wrap the slider with a React component or apply other frameworks.
 
 Mouseover and click to view `:hover` and `:active` state.
+
+To disabled state, add the `a-slider--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
 
 <Playground>
   <div className="a-slider">
--- a/docs/controls-toggles.mdx
+++ b/docs/controls-toggles.mdx
@@ -121,19 +121,19 @@ Mouseover and click to view `:hover` and `:active` state.
 
 <Playground>
   <div className="a-slider">
-    <input type="range" min="1" max="100" id="slider1" />
     <label htmlFor="slider1">
       Slider
       <span className="a-slider--value">50 coins</span>
     </label>
+    <input type="range" min="1" max="100" id="slider1" />
   </div>
 
-  <div className="a-slider">
-    <input type="range" min="1" max="100" id="slider1" disabled />
+  <div className="a-slider a-slider--disabled">
     <label htmlFor="slider1">
       Disabled
       <span className="a-slider--value">0 coins</span>
     </label>
+    <input type="range" min="1" max="100" id="slider1" disabled />
   </div>
 </Playground>
 
--- a/docs/controls-toggles.mdx
+++ b/docs/controls-toggles.mdx
@@ -20,39 +20,39 @@ Mouseover to view `:hover` state.
 
 <Playground>
   <div className="a-radio">
-    <input type="radio" id="radio1" name="radiogroup1" />
     <label htmlFor="radio1">Radio</label>
+    <input type="radio" id="radio1" name="radiogroup1" />
   </div>
 
-<div className='a-radio'>
-  <input type='radio' id='radio2' name='radiogroup1' defaultChecked />
-  <label htmlFor='radio2'>Selected</label>
-</div>
+  <div className="a-radio">
+    <label htmlFor="radio2">Selected</label>
+    <input type="radio" id="radio2" name="radiogroup1" defaultChecked />
+  </div>
 
   <div className="a-radio a-radio--disabled">
-    <input type="radio" id="radio3" name="radiogroup1" disabled />
     <label htmlFor="radio3">Disabled</label>
+    <input type="radio" id="radio3" name="radiogroup1" disabled />
   </div>
 </Playground>
 
 Default label behavior is top alignment. You can add the `a-radio--align-middle` modifier class to switch to middle alignment.
 
 <Playground>
-  <div className='a-radio'>
-    <input type='radio' id='radio4' name='radiogroup2' />
-    <label htmlFor='radio4'>
+  <div className="a-radio">
+    <label htmlFor="radio4">
       Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
       tempor incididunt ut labore et dolore magna aliqua. Sed risus pretium quam
       vulputate dignissim suspendisse.
     </label>
+    <input type="radio" id="radio4" name="radiogroup2" />
   </div>
-  <div className='a-radio a-radio--align-middle'>
-    <input type='radio' id='radio5' name='radiogroup2' />
-    <label htmlFor='radio5'>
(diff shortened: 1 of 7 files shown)


src/css/radio.css
```diff
@@ -1,67 +1,82 @@
...
   border-color: var(--color-moon-200);
 }
 
-/* Hover state */
+/* Cursor */
 
-.a-radio > input[type='radio']:hover:not(:disabled) + label::before {
-  border-color: var(--color-uranus-500);
+.a-radio:not(.a-radio--disabled) > label,
+.a-radio:not(.a-radio--disabled) > input[type='radio']::before {
```
luisfmsouza: What is the behaviour of the cursor when the input has the property `disabled` but don't have the class?

docs/inputs.mdx
```diff
@@ -20,29 +20,29 @@ The floating label will need JS to work on your project. You may find the code f
 
 <Playground>
   <div className="a-input">      
-    <input type="text"/>
-    <label>Standard</label>
+    <input id="input1" type="text" aria-labelledby="label1" />
+    <label id="label1" htmlFor="input1">Standard</label>
```
luisfmsouza: I think we should invert the input and label order, dont we?

src/css/radio.css
brunobc182: Inputs still disabled and cursor is not allowed. I did some changes here 0c982c208cc0ff3a6dc8bf0944d8240496b4a489.

docs/inputs.mdx
brunobc182: Yes, but we discussed and we came with a conclusion that our CSS is too dependent on HTML structure. So we tested those attributes with screen reader and get the same result with the label above inputs. 

What do you  think?

Auralcat:
[💻: 3, Padronizar estado `disabled` dos componentes](https://trello.com/c/mzvBLaKq/389-%F0%9F%92%BB-3-padronizar-estado-disabled-dos-componentes)

docs/inputs.mdx
luisfmsouza: Just to complement: what's the exactly output we got from screen readers?

brunobc182: To test we used the [Chrome Vox extension](https://chrome.google.com/webstore/detail/chromevox/kgejglhpjiefppelpmljglcjbhoiplfn/related?hl=pt-BR)  and the output we select the input was the name 
of the label, in that case `Standard` then input type.

brunobc182:
> Why are we disabling this rule?

I saw in another file and decided to add to this one. Its allows us to group classes by type. But honestly, I do not like to override the rules of lint. What do you think?

docs/controls-toggles.mdx
```diff
@@ -16,41 +16,43 @@ Please follow the markup structure from the examples below. You must create a un
 
 Mouseover to view `:hover` state.
 
+To disabled state, add the `a-radio--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
```
larimaza:
```suggestion
To create the disabled state, add the `a-radio--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
```

docs/controls-toggles.mdx
```diff
@@ -60,25 +62,27 @@ Checkboxes are used when a user might select multiple options from a list, or wh
 
 Mouseover to view `:hover` state.
 
+To disabled state, add the `a-checkbox--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
```
larimaza:
```suggestion
To create the disabled state, add the `a-checkbox--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
```

docs/controls-toggles.mdx
```diff
@@ -88,6 +92,8 @@ A simple control that is used to quickly switch between two possible states.
 
 Mouseover to view `:hover` state.
 
+To disabled state, add the `a-toggle--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
```
larimaza:
```suggestion
To create the disabled state, add the `a-toggle--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
```

docs/controls-toggles.mdx
```diff
@@ -117,21 +123,23 @@ The colored percentage fill and counter label will need JS to work on your proje
 
 Mouseover and click to view `:hover` and `:active` state.
 
+To disabled state, add the `a-slider--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
```
larimaza:
```suggestion
To create the disabled state, add the `a-slider--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
```

docs/inputs.mdx
```diff
@@ -17,31 +17,33 @@ Please follow the markup structure below in the same order to get all correct CS
 
 The floating label will need JS to work on your project. You may find the code for this working example in our GitHub repository.
 
+To disabled state, add the `a-input--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
```
larimaza:
```suggestion
To create the disabled state, add the `a-input--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
```

docs/inputs.mdx
larimaza: If that's the case I suppose it's ok to keep this structure...

larimaza:
[Padronizar estado `disabled` dos componentes](https://trello.com/c/HA8HJv9w/105-padronizar-estado-disabled-dos-componentes)

larimaza:
[Padronizar estado `disabled` dos componentes](https://trello.com/c/HA8HJv9w/105-padronizar-estado-disabled-dos-componentes)

docs/inputs.mdx
luisfmsouza: Agree!

src/css/radio.css
```diff
@@ -1,67 +1,80 @@
-/* Hiding standard input */
+/* stylelint-disable no-descending-specificity */
```
luisfmsouza: [resuming this discussion](https://github.com/magnetis/astro/pull/51#commitcomment-32683503)

There is a good reason for [the rule `no-descending-specificity` existence](https://stylelint.io/user-guide/rules/no-descending-specificity/):

> If a selector with higher specificity comes before the selector it overrides, we have to think harder to understand it, because it violates the source order expectation. Stylesheets are most legible when overriding selectors always come after the selectors they override. That way both mechanisms, source order and specificity, work together nicely.

I think we should rethink if we are going to take this rule on our favor, or not. I mean, if we don't agree with it, think we should revoke it for the whole repo. If we are going to use this rule, we can rethink our organisation. Let's bring the linter to our side!

cc @brunobc182 @larimaza

brunobc182: IMHO, I think we should just let him do his job. As I said earlier, I do not like to disable rules.
What do you guys think?

cc @elmarques @larimaza

luisfmsouza: @brunobc182 if that's going to be our choice, my suggestion for this PR is remove the disable comment from the file we are adding it and adjust the code to respect the rule :)

elmarques:
Thanks for finding these bugs @larimaza! :pray: We fixed the wrong radio spacing with this commit (ccc1520), and we also reviewed the hover area bug here (9489e96).

I think the cursor is pretty stable now. Could you test again, please?

src/css/radio.css
elmarques: I found more comments like this on other components (button, checkbox, radio, slider, and toggle). It seems to be a project need.

a) We can remove the disabler comments from all CSS files above and add the `no-descending-specificity-rule: null` line in `.stylelintrc` config to make a project scope decision;

b) We can fix this rule for this PR's files and create a tech debt to fix the others;

c) We can merge this PR as it is and then decide whether to use the rule or not.

We need to evaluate the use of this specific rule. Maybe it can disrupt the way we organize the code (we usually divide the component by sections: base, designs, states, etc.).

I think it's not a sin to stop using specific rules as long as you have good reasons, and keeping the code more organized and easy to maintain are good ones for me.

What do you think @larimaza?

kuroski: I agree with the `b/c` options.

I think this rule is important, and maybe for now we can merge this PR and use something like:
- https://stylelint.io/user-guide/cli/#autofixing-errors
- http://csscomb.com/
- https://css-tricks.com/prettier-stylelint-writing-clean-css-keeping-clean-code-two-tool-game/

To prevent us from doing lint errors and also fix that for us automatically.

We can use the autofix to run every commit or every build task or even fix errors on save.
And linters are awesome, but they also distract us from doing our task if the error don't get fixed automatically.

elmarques: In the meanwhile, we fixed the linter warnings for checkbox, radio and toggle components here (a37d296).

The buttons and slider will require more time/work.

Do you think we can merge this PR and open a tech debt to finish this job later?

kuroski: I think it's a nice approach =D

brunobc182: We can configure https://github.com/typicode/husky to run a hook on git push. This way it is easy to catch lint errors.

larimaza: Sorry about the delay! I think these selectors make sense together for later maintenance. That's why I disabled the rule. I vote for merging this PR and tackling that discussion later with a tech debt that would fix the whole project. :)

larimaza: https://trello.com/c/gwXRtNA0/120-rever-regra-de-linter-no-projeto-no-descending-specificity

PR #58, opened 2019-03-14
Fix single quote formatting in .mdx files

larimaza:
# What

Fix the formatting of .mdx files in the project from inconsistent double/single quotes to all-single quotes.

# Why

Prettier wasn't able to format all .mdx files correctly when there's an empty paragraph between elements inside a `<Playground>`.

# How

Remove all empty paragraphs between elements.


larimaza:
[Corrigir formatação dos arquivos .mdx](https://trello.com/c/WroFHdeo/118-corrigir-formata%C3%A7%C3%A3o-dos-arquivos-mdx)

PR #60, opened 2019-03-18
Create text area input classes and components

Auralcat:
# What

This PR creates the styles for textarea components on docz.

# Why

Textarea elements are important in form inputs to get longer kinds of text from the user.
Their style should be consistent with the design system.

# How

- Create CSS file for textarea tag classes
- Add textarea component playground in `inputs.mdx`

# Sample

Screenshot

### DIFF
--- a/docs/inputs.mdx
+++ b/docs/inputs.mdx
@@ -68,4 +68,20 @@ To create the disabled state, add the `a-input--disabled` modifier class on elem
   </div>
 </Playground>
 
+## text area inputs
+
+Text areas are used in contexts that demand long and descriptive texts.
+
+<Playground>
+  <div className='a-input'>
+    <textarea name='random' id='textarea1' rows='10' cols='20'></textarea>
+    <label id='label6' htmlFor='textarea1'>Write your book here</label>
+  </div>
+
+  <div className='a-input a-input--disabled'>
+    <textarea name='random' id='textarea2' rows='10' cols='20' disabled></textarea>
+    <label id='label7' htmlFor='textarea2'>Disabled</label>
+  </div>
+</Playground>
+
 <HandleFloatingLabel />
--- a/docs/js/inputs.js
+++ b/docs/js/inputs.js
@@ -11,7 +11,7 @@ export const initFloatingLabel = () => {
     };
 
     inputs.forEach(input => {
-      const currentInput = input.querySelector("input");
+      const currentInput = input.firstChild;
       const currentLabel = input.querySelector("label");
 
       verifyValue(currentInput, currentLabel);
--- a/src/css/inputs.css
+++ b/src/css/inputs.css
@@ -3,7 +3,8 @@
   display: inline-block;
 }
 
-.a-input > input {
+.a-input > input,
+.a-input > textarea {
   width: 100%;
   padding: 25px 44px 5px 16px;
   border: 1px solid var(--color-moon-500);
@@ -39,11 +40,13 @@
   color: var(--color-mars-500);
 }
 
-.a-input > input:focus {
+.a-input > input:focus,
+.a-input > textarea:focus {
   border: 1px solid var(--color-uranus-500);
 }
 
-.a-input > input:focus + label {
+.a-input > input:focus + label,
+.a-input > textarea:focus + label {
   top: 8px;
   color: var(--color-uranus-500);
   font-size: 12px;
@@ -103,13 +106,15 @@
   font-size: 12px;
 }
 
-.a-input--disabled > input:disabled {
+.a-input--disabled > input:disabled,
+.a-input--disabled > textarea:disabled {
   cursor: not-allowed;
   border: 1px solid var(--color-moon-200);
   background-color: var(--color-moon-100);
 }
 
-.a-input--disabled > input:disabled + label {
+.a-input--disabled > input:disabled + label,
+.a-input--disabled > textarea:disabled + label {
   cursor: not-allowed;
   color: var(--color-moon-300);
   font: var(--font-secondary);


Auralcat:
[Criar componentes de "text area input" e inserir na página "Inputs" na documentação](https://trello.com/c/bRN0u8V0/109-criar-componentes-de-text-area-input-e-inserir-na-p%C3%A1gina-inputs-na-documenta%C3%A7%C3%A3o)

src/css/inputs-textarea.css
```diff
@@ -0,0 +1,116 @@
+.a-textarea {
```
brunobc182: I think you could just name this file as `textarea.css`. It's already implicit that a input. :grimacing:

docs/inputs.mdx
```diff
@@ -68,4 +68,20 @@ To create the disabled state, add the `a-input--disabled` modifier class on elem
...
+
+Text areas are used in contexts that demand long and descriptive texts.
+
+<Playground>
+    <div class='a-textarea'>
+        <textarea name='random' id='' rows='10' cols='20' tabindex=''></textarea>
+        <label>Write your book here</label>
+    </div>
+
+    <div class='a-textarea'>
```
brunobc182: I think you need to put `a-textarea--disabled` to apply the correct style on disabled state.

src/css/inputs-textarea.css
```diff
@@ -0,0 +1,116 @@
...
+}
+
+.a-textarea > textarea:focus + label {
+  top: 8px;
+  color: var(--color-uranus-500);
+  font-size: 12px;
+  font-weight: 700;
+}
+
+.a-textarea > .a-textarea--floating-label {
```
brunobc182: The floating label will not work, because on  [input.jsf](https://github.com/magnetis/astro/blob/master/docs/js/inputs.js) file we only handle `a-input--floating-label` class. I think we can change it to a generic name to work on every input. 

What do you guys think? :thinking:

docs/inputs.mdx
```diff
@@ -68,4 +68,20 @@ To create the disabled state, add the `a-input--disabled` modifier class on elem
   </div>
 </Playground>
 
+## text area inputs
+
+Text areas are used in contexts that demand long and descriptive texts.
+
+<Playground>
+    <div class='a-textarea'>
+        <textarea name='random' id='' rows='10' cols='20' tabindex=''></textarea>
```
brunobc182: I think you should use `id` on textarea and `htmlFor` on label for purposes of accessibility.

src/css/inputs-textarea.css
Auralcat: Changed in https://github.com/magnetis/astro/commit/a5dddb06f811326f70305e91b019b393efb17d76

docs/inputs.mdx
Auralcat: Done in https://github.com/magnetis/astro/commit/a5dddb06f811326f70305e91b019b393efb17d76

docs/inputs.mdx
Auralcat: Done in https://github.com/magnetis/astro/commit/a5dddb06f811326f70305e91b019b393efb17d76.

BTW, are the `aria-labelledby` attributes necessary for textareas as well?

brunobc182: I tested only with `id` and `htmlFor` without `aria-labelledby` and the screen read handled it. I think it's not really necessary.

src/css/inputs-textarea.css
larimaza: Actually... The `inputs.css` file already contains two kinds of input (one of them is in this open PR: https://github.com/magnetis/astro/pull/62/files), so we should keep all of them in that same file! 🙈

larimaza: And, since they'll be in the same file, it would be best if the class was named like the standard `a-input--textarea` :)

src/css/inputs-textarea.css
larimaza: Will the CSS still work if we change this class name to `a-input--floating-label` to get all of the same stuff?

src/css/inputs-textarea.css
Auralcat: Done in https://github.com/magnetis/astro/commit/fe472b2dc66b38a48f74e1dfc2f44294905accce

src/css/inputs-textarea.css
brunobc182: My  proposal is, add `a-input` and `a-input--textarea` also change the `input.js` file.

- `input.mdx`: 

```html
<div className='a-input a-input--textarea'>
  <textarea name='random' id='textarea1' aria-labelledby='label6' rows='10' cols='20' tabIndex=''></textarea>
  <label id='label6' htmlFor='textarea1'>Write your book here</label>
</div>
```

- `input.js`

Change this on `line 14`:
```javascript
const currentInput = input.querySelector("input");
```

To:

```javascript
const currentInput = input.querySelector('input') ? input.querySelector('input') : input.querySelector('textarea');
```

I think it's covers the two cases of floating label. If we have a better approach, please let me know. :smile:

Auralcat:
[ch2038]

src/css/inputs-textarea.css
larimaza: I honestly don't love the idea of having to use two classes if we can avoid this...
Maybe there's a way to use `querySelector` to select all elements _containing_ "input" - and since all inputs contain "input" in their classes now, it should work... 
What do you think?

Auralcat: We could use the `input` and `textarea` classes both as `.a-input` since the style is implemented in its children. The selectors are like this: `.a-input > textarea` or `.a-input > input`.

The drawback of this approach is that we won't have a specific class for the textarea component. WDYT?

brunobc182: Since our HTML struct needs to follow the right way to work. I think we just need to take the first element inside `.a-input` class. For example: `const currentInput = input.firstChild;` It's should work for both cases.

Auralcat: Implemented `firstChild` selection in the project + minor refactors in https://github.com/magnetis/astro/commit/c41f2dbbc6c20d2491a6e6d90a823dee98b1e5e9

brunobc182:
Just one thing. I think you don't need the `src/css/textarea.css` file. :)

Auralcat:
Removed the extra file in https://github.com/magnetis/astro/commit/c41f2dbbc6c20d2491a6e6d90a823dee98b1e5e9

docs/inputs.mdx
```diff
@@ -68,4 +68,20 @@ To create the disabled state, add the `a-input--disabled` modifier class on elem
...
+Text areas are used in contexts that demand long and descriptive texts.
+
+<Playground>
+  <div className='a-input'>
+    <textarea name='random' id='textarea1' aria-labelledby='label6' rows='10' cols='20'></textarea>
+    <label id='label6' htmlFor='textarea1'>Write your book here</label>
+  </div>
+
+  <div className='a-input a-input--disabled'>
+    <textarea name='random' id='textarea2' aria-labelledby='label7' rows='10' cols='20' disabled></textarea>
```
larimaza: What was settled here - was this aria label going to be removed? 🤔

Auralcat: Solved in https://github.com/magnetis/astro/commit/1bedc8e0f9364aa22ac475313d1c227571765db3

PR #64, opened 2019-03-26
Add icon buttons to src CSS and documentation

larimaza:
# What

* Create all types of icon buttons on src CSS file and reflect them in the documentation;
* Adjust icons CSS to better reflect the specs.

# Why

This PR adds icon buttons to Astro according to design specs.

# How

* Change `buttons.css` file to create variations for the icon buttons and make them work with any combination of button style, state, color and size. That's quite a complex array of combinations, so I had to adjust the order of some declarations in order to get the most of the cascade, without resorting to specificity hacks.
* Add samples of the icon buttons to the `buttons.mdx` doc file.
* Adjust icon size modifiers to better reflect current use cases, such as the buttons here displayed.
* Fix english typo in balloon icon that said "baloon".

# Sample

As there are many buttons, please check the staging URL in the bottom of this page under "Checks" :)


larimaza:
[Criar os botões com ícones no CSS e documentação](https://trello.com/c/7T0zaPLD/89-criar-os-bot%C3%B5es-com-%C3%ADcones-no-css-e-documenta%C3%A7%C3%A3o)

src/css/icons.css
```diff
@@ -6,7 +6,7 @@
   height: 1em;
   vertical-align: middle;
   background-color: var(--color-moon-900);
-  font-size: 24px;
+  font-size: 32px;
```
elmarques: 🙏 Thanks!

docs/buttons.mdx
```diff
@@ -77,10 +77,110 @@ Mouseover and click to view `:hover` and `:active` states.
   </button>
 </Playground>
 
+## iconlabel buttons
```
elmarques: Just a useless comment: don't you think this *iconlabel* expression (without space) weird?

docs/buttons.mdx
```diff
@@ -77,10 +77,110 @@ Mouseover and click to view `:hover` and `:active` states.
...
+  <button
+    className='a-btn a-btn--ghost-uranus a-btn--medium a-btn--iconlabel a-btn--iconlabel--right'
+    disabled
+  >
+    iconlabel
+    <i class='a-icon a-icon__calendar a-icon--size-medium' />
+  </button>
+</Playground>
+
+## icon buttons
```
elmarques: Just to honor the OCD Goddess 👩‍🎨, do you think we should change the subheadings **primary** to **primary buttons** and **outline** to **outline buttons** since all other subtitles follow this pattern _(... buttons)_?

docs/buttons.mdx
```diff
@@ -111,6 +211,63 @@ Buttons can have different sizes, also by changing their CSS size class.
   </button>
 </Playground>
 
+<Playground>
```
elmarques: Great decision to group the examples in separate playgrounds. It was different from the layout, but I would do the same. 👍

docs/iconography.mdx
```diff
@@ -66,7 +66,7 @@ All icons have a 32x32px bound as default format. They can be resized in our 16p
     'pizza',
     'circle-mail',
     'configuration',
-    'baloon',
+    'balloon',
```
elmarques: 👼

src/css/buttons.css
```diff
@@ -288,4 +267,167 @@
   color: var(--color-moon-200);
 }
 
+/* Iconlabel button */
+
+.a-btn--iconlabel .a-icon {
```
elmarques: If this expression is composed then it's ok.

Otherwise, BEM supports something like:

```suggestion
.a-btn--icon-label .a-icon {
```

src/css/buttons.css
```diff
@@ -288,4 +267,167 @@
...
+  background-color: var(--color-space-100);
+  margin: -4px 9px -3px 0;
+  transition: all 0.3s ease;
+}
+
+.a-btn--iconlabel--right .a-icon {
```
elmarques: I think for compound modifiers, BEM indicates something like:

```suggestion
.a-btn--iconlabel-right .a-icon {
```

docs/buttons.mdx
larimaza: I do 😬 will check with design folks if it was intended!

docs/buttons.mdx
larimaza: LOL I'm glad the goddess is a woman! I confess I thought about that but it went away, thanks for the reminder.

src/css/buttons.css
larimaza: If we're going to keep "iconlabel" with no space, I think we should keep that...

docs/buttons.mdx
larimaza: They confirm it's a neologism, so it's ok to leave it that way, I guess

src/css/buttons.css
larimaza: Done on 6a35751

docs/buttons.mdx
larimaza: Done on 6a35751

PR #66, opened 2019-04-03
Make new theme responsive

larimaza:
# What

Make adjustments so that the new theme doesn't break on smaller devices.

# Why

We might have mobile users.

# How

- Adjust CSS in themeConfig in `doczrc.js` and in the CSS file for the home components.
- Had to create a hard coded solution for the flex containers in the home to turn responsive, because of the stubborn code snippet which needs a defined width (else it takes up too much space).

# Unrelated

- Adjust other files that said "class" instead of "className", noticed thanks to the dev console which pointed the errors out.

# Test

Access the staging URL under "Checks" in the bottom of this page :)

### DIFF
--- a/docs/css/home.css
+++ b/docs/css/home.css
@@ -2,15 +2,32 @@
 
 .cards-box {
   display: flex;
+  flex-wrap: wrap;
   justify-content: space-between;
-  margin-bottom: 50px;
 }
 
 .cards-box > div {
   display: flex;
+  flex: 1;
   flex-direction: column;
   justify-content: space-between;
-  width: 30%;
+  margin-bottom: 20px;
+  padding-right: 5%;
+  width: 100%;
+}
+
+@media (min-width: 768px) {
+
+  .cards-box > div {
+    width: 50%;
+  }
+}
+
+@media (min-width: 865px) {
+
+  .cards-box > div {
+    width: 30%;
+  }
 }
 
 .cards-box .a-btn {
--- a/doczrc.js
+++ b/doczrc.js
@@ -9,6 +9,10 @@ const selectors = {
   menuItem: '& .css-82rrzo, .css-82rrzo.active, .css-82rrzo:visited',
   menuSubitemsArea: '& .css-1qc5yjq',
   menuSubItem: '& .css-1l1jc95, .css-1l1jc95:visited',
+  menuBtn: '& .css-6pue65, .css-15exo54',
+  menuBtnShape: '& .css-14xo6fr, .css-1v3qlja',
+  mobileLogoArea: '& .css-12xm67f',
+  mobileMenu: '& .css-k33w14',
 
   /* Docz default elements (hidden) */
 
@@ -129,6 +133,11 @@ export default {
             backgroundColor: 'var(--color-moon-900)'
           }
         },
+        '@media (max-width: 1119px)': {
+          [selectors.menuItem]: {
+            padding: '22px 39px',
+          }
+        },
         [selectors.menuSubitemsArea]: {
           margin: '5px 0 15px'
         },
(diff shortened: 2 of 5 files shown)


larimaza:
[Ajustar o comportamento responsivo do tema novo do Docz](https://trello.com/c/By4n95zu/122-ajustar-o-comportamento-responsivo-do-tema-novo-do-docz)

doczrc.js
```diff
@@ -148,18 +157,45 @@ export default {
...
+          background: 'transparent',
+          '&:before': {
+            display: 'none'
+          }
+        },
+        [selectors.menuBtnShape]: {
+          height: 0,
+          border: '1px solid var(--color-space-100)',
+          borderRadius: 3,
+          background: 'var(--color-space-100)'
```
elmarques:
```suggestion
          backgroundColor: 'var(--color-space-100)'
```

💅 👩‍🎨

doczrc.js
```diff
@@ -148,18 +157,45 @@ export default {
...
+        },
+        [selectors.mobileLogoArea]: {
+          padding: 0,
+          '&:before': {
+            display: 'none'
+          }
+        },
+        [selectors.mobileMenu]: {
+          background: 'var(--color-space-100)'
```
elmarques:
```suggestion
          backgroundColor: 'var(--color-space-100)'
```
💅 👩‍🎨

doczrc.js
```diff
@@ -148,18 +157,45 @@ export default {
           '&:before': {
             display: 'none'
           }
-        }
+        },
+        [selectors.menuBtn]: {
+          background: 'transparent',
```
elmarques:
```suggestion
          backgroundColor: 'transparent',
```
💅 👩‍🎨

doczrc.js
```diff
@@ -201,6 +237,7 @@ export default {
       },
       playground: {
         padding: 16,
+        overflowX: 'scroll',
```
elmarques: Can I ask you two things in this `doczrc.js` file, please?

1. Put this comment on line 33 for the style-lint stop complaining:
`preprocessor: 'postcss' /* stylelint-disable-line */`

2. Run Prettier formatter to delete some extra commas

Sorry for the OCD. 😁

docs/css/home.css
```diff
@@ -2,15 +2,32 @@
...
+  flex: 1;
   flex-direction: column;
   justify-content: space-between;
-  width: 30%;
+  margin-bottom: 20px;
+  padding-right: 5%;
+  width: 100%;
+}
+
+@media (min-width: 768px) {
```
elmarques: This made me wonder if it would be nice to have a breakpoint variables file in the future. 🤔 

If so, we'll need to install a plugin to support CSS variables in media queries](https://www.npmjs.com/package/postcss-media-variables).

But I think we can chat before to align this point. What do you think?

doczrc.js
larimaza: No idea why I thought I had to use "background" shorthand to override "background" from Docz styles, apparently not 😅

doczrc.js
larimaza: Sure thing! eb83eeb

doczrc.js
larimaza: eb83eeb

doczrc.js
larimaza: eb83eeb

doczrc.js
larimaza: eb83eeb

docs/css/home.css
larimaza: I believe that's not in the Astro scope yet, as we only store components and breakpoints are related to the page layout 🤔 
As for the theme, I think we don't need a scalable solution, being a one-time thing...

PR #68, opened 2019-04-10
Refactor Controls & Toggles to fix pseudoclass bug

larimaza:
# What

Refactor radio, checkbox, toggle and slider components in Controls & Toggles to make them work in all browsers.

# Why

There was an issue report stating that these components did not work in Firefox, because only Chrome accepts pseudoclasses on `input` tags.

# How

* Remove styles from the `input` tag, using it only to handle behavior now;
* Create a new sibling `div` to draw the component shape (this must be after the `input` in the markup order, to make selection possible when handling states);
* Adjust CSS accordingly.

### DIFF
--- a/docs/controls-toggles.mdx
+++ b/docs/controls-toggles.mdx
@@ -95,17 +95,17 @@ To create the disabled state, add the `a-toggle--disabled` modifier class on ele
   <div className='a-toggle'>
     <label htmlFor='toggle1'>off</label>
     <input type='checkbox' id='toggle1' />
-    <div className='a-toggle--shape' />
+    <span className='a-toggle__shape' />
   </div>
   <div className='a-toggle'>
     <label htmlFor='toggle2'>on</label>
     <input type='checkbox' id='toggle2' defaultChecked />
-    <div className='a-toggle--shape' />
+    <span className='a-toggle__shape' />
   </div>
   <div className='a-toggle a-toggle--disabled'>
     <label htmlFor='toggle3'>disabled</label>
     <input type='checkbox' id='toggle3' disabled />
-    <div className='a-toggle--shape' />
+    <span className='a-toggle__shape' />
   </div>
 </Playground>
 
--- a/docs/controls-toggles.mdx
+++ b/docs/controls-toggles.mdx
@@ -73,18 +73,22 @@ To create the disabled state, add the `a-checkbox--disabled` modifier class on e
   <div className='a-checkbox'>
     <label htmlFor='checkbox1'>Checkbox</label>
     <input type='checkbox' id='checkbox1' />
+    <span className='a-checkbox__shape' />
   </div>
   <div className='a-checkbox'>
     <label htmlFor='checkbox2'>Checked</label>
     <input type='checkbox' id='checkbox2' defaultChecked />
+    <span className='a-checkbox__shape' />
   </div>
   <div className='a-checkbox a-checkbox--indeterminate'>
     <label htmlFor='checkbox3'>Indeterminate</label>
     <input type='checkbox' id='checkbox3' defaultChecked />
+    <span className='a-checkbox__shape' />
   </div>
   <div className='a-checkbox a-checkbox--disabled'>
     <label htmlFor='checkbox4'>Disabled</label>
     <input type='checkbox' id='checkbox4' disabled />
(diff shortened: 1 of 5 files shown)


larimaza:
[Refatorar componentes em Controls & Toggles para corrigir bug no Firefox](https://trello.com/c/Uls6HN9z/125-refatorar-componentes-em-controls-toggles-para-corrigir-bug-no-firefox)

src/css/toggle.css
```diff
@@ -6,30 +6,41 @@
...
+  z-index: var(--z-index-1);
 }
 
 /* Label */
 
-.a-toggle > label {
+.a-toggle label {
   padding-right: 8px;
   font: var(--font-secondary);
+  float: left;
```
elmarques:
```suggestion
```

A suggestion in the next comment.

src/css/toggle.css
```diff
@@ -6,30 +6,41 @@
...
 }
 
 /* Base */
 
-.a-toggle > input[type='checkbox']::before {
+.a-toggle--shape {
+  position: relative;
+  float: left;
```
elmarques: I don't recommend using `float` (even more without the `clear`). This can be tricky for the next layout element [[1]](https://www.grandcircus.co/blog/why-ive-stopped-using-float-in-my-css/) [[2]](https://thoughtbot.com/blog/how-i-learned-to-stop-floating-and-love-the).

As a possible alternative, maybe we can use:

```css
.a-toggle__shape {
  position: relative;
  cursor: pointer;
  display: inline-block;
  vertical-align: middle;
}
```

What do you think?

src/css/toggle.css
```diff
@@ -56,42 +67,35 @@
 
 /* On state */
 
-.a-toggle > input[type='checkbox']:checked::before {
```
elmarques: Are we going to remove this direct children selector ` > `?

If I'm not mistaken, several other Astro components use this approach (I think it was Kuroski's recommendation).

In my opinion, I prefer without `>` like you did 😸 but I just think we should standardize it.

docs/controls-toggles.mdx
```diff
@@ -95,14 +95,17 @@ To create the disabled state, add the `a-toggle--disabled` modifier class on ele
   <div className='a-toggle'>
     <label htmlFor='toggle1'>off</label>
     <input type='checkbox' id='toggle1' />
+    <div className='a-toggle--shape' />
```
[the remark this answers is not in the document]
larimaza: Huh, I thought `div` was more of a regular practice to draw shapes... But we can definitely change that :) Why do you say it's more semantic?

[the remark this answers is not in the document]
larimaza: True!

src/css/toggle.css
larimaza: We had removed it because our first try was wrapping the input with the shape div (and thus the direct selector would not work). We can put it back :) I agree with using > since it's already our standard.

src/css/toggle.css
larimaza: Wow. Still works, so that's a yes!

docs/controls-toggles.mdx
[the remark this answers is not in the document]
elmarques: I think your point also makes sense. I don't have a strong position in this.

Perhaps this excerpt best illustrates my suggestion:

> The difference between span and div is that a span element is in-line and usually used for a small chunk of HTML inside a line (such as inside a paragraph) whereas a div (division) element is block-line (which is basically equivalent to having a line-break before and after it) and used to group larger chunks of code.

https://htmldog.com/guides/html/intermediate/spandiv/

Sorry for the lack of context. 🤓

src/css/toggle.css
larimaza: Done on fa44c47

src/css/toggle.css
larimaza: Done on fa44c47

docs/controls-toggles.mdx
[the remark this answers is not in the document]
larimaza: Done on fa44c47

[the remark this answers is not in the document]
larimaza: Looks like there isn't muuuch of a difference 🤔 But I did it anyways hahah Done on fa44c47

PR #75, opened 2019-04-22
Fix messaging input

brunobc182:
# What

This **PR** change docs and the default button color of the messaging input. Also replaces the `arrow-right.svg` with a new one with the correct centering.

# How

* Change button color and docs text
* Add new arrow-right.svg with correct centering

# Sample

Screenshot from 2019-04-22 11-26-13


brunobc182:
[ch5127]

brunobc182:
[Atualizar componente de messaging input](https://trello.com/c/utE0YA1N/124-atualizar-componente-de-messaging-input)

PR #76, opened 2019-04-22
Fix input label color inconsistency

larimaza:
# What

Fix input label color inconsistency.

# Why

In our docs, input label color is always moon 300. But according to the specs, when the label floats, it should change to moon 500.

# How

As there's a complex conflict between the need to set a color for the floating label and the need to keep the invalid label red at all times (selector specificity was getting crazy), we agreed with the design team to set one standard grey for all states.
The label is now moon 400 when resting _and_ floating.

# Sample

**Before** | **After**
-|-
Screen Shot 2019-04-22 at 14 14 27|Screen Shot 2019-04-22 at 14 14 08

### DIFF
--- a/src/css/inputs.css
+++ b/src/css/inputs.css
@@ -21,12 +21,12 @@
   left: 16px;
   transition: 0.2s ease all;
   pointer-events: none;
-  color: var(--color-moon-300);
+  color: var(--color-moon-400);
   font: var(--font-secondary);
 }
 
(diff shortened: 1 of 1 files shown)


larimaza:
[Corrigir cores das labels flutuantes dos inputs](https://trello.com/c/RYQjdH9M/126-corrigir-cores-das-labels-flutuantes-dos-inputs)

PR #79, opened 2019-04-25
Create ghost inputs and add them to Inputs

larimaza:
# What

Create ghost input components and add them to the Inputs page on the docs.

# Why

We have design specs for that input style.

# How

- Create styles for the ghost inputs in `inputs.css`;
- Create a class to add a gradient background to the Playground component in the themeConfig object in `doczrc.js` file;
- Add a sample of the component in the Inputs doc page;
- Absolutely had to disable the Stylelint rule for the CSS file, else I'd have to scatter the ghost input styles in different places in the file. Will reorganize the whole file later, since it's all scattered because of that rule 😬 

# Sample

Screen Shot 2019-04-25 at 16 22 02

### DIFF
--- a/docs/inputs.mdx
+++ b/docs/inputs.mdx
@@ -86,6 +86,32 @@ Text areas are used in contexts that demand long and descriptive texts.
   </div>
 </Playground>
 
+## ghost inputs
+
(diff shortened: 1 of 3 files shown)


larimaza:
[Criar componentes de "ghost input" e inserir na página "Inputs" na documentação](https://trello.com/c/i2XXOrwL/127-criar-componentes-de-ghost-input-e-inserir-na-p%C3%A1gina-inputs-na-documenta%C3%A7%C3%A3o)

PR #80, opened 2019-04-25
Create ghost icon buttons and reorganize all icon buttons

larimaza:
# What

Create the variant of ghost icon buttons in the documentation and reorganize icon butons, separating each variant.

# Why

We are going to need ghost icon buttons too.

# How

- Combine ghost button class with icon buttons to create the new ghost icon button variant - no need for new CSS <3
- Reorganize the components to create separate sections for regular icon buttons, outline icon buttons and ghost icon buttons.
- Include all colors to make the docs more interesting 😏 

# Sample

Screen Shot 2019-04-25 at 19 09 30

### DIFF
--- a/docs/buttons.mdx
+++ b/docs/buttons.mdx
@@ -155,7 +155,7 @@ Ghost iconlabel buttons are also used in microinteractions and smaller commands.
 
 Icon buttons can be used for simple microinteractions and commands.
 
-To create an icon button, first create a regular primary or outline button with all size and color modifier classes, plus the `a-btn--icon` class. Then add a `<i>` tag inside the button with no inner text.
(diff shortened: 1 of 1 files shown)


larimaza:
[Criar componentes de "ghost icon buttons" e inserir na página "Buttons" na documentação](https://trello.com/c/g52yGKFl/132-criar-componentes-de-ghost-icon-buttons-e-inserir-na-p%C3%A1gina-buttons-na-documenta%C3%A7%C3%A3o)

Auralcat:
I noticed that when you hover over the ghost button, the text inside goes from the earth color to a lighter tone, is that the expected behavior? Just curious, because the other buttons retain the same text color when hovering.

larimaza:
> I noticed that when you hover over the ghost button, the text inside goes from the earth color to a lighter tone, is that the expected behavior? Just curious, because the other buttons retain the same text color when hovering.

It is, when regular ghost buttons (already live in the docs) are hovered this happens too :)

PR #84, opened 2019-04-30
Make color variables accessible in the bundle

larimaza:
# What

Add explanation on how to import colors and gradients variables to Colors documentation.

# Why

Astro users might need to access these variables, and we found importing them directly was the best option for now, as our compiler kills them in the bundle.

# How

Just add some copy to the Colors page.

# QA

Test the method described here and see if the variables work.

# Update

We found that there's a config option to preserve the variables in the bundle while still converting them to hexadecimal in the browsers, so that's what we're doing now in this PR.

### DIFF
--- a/docs/colors.mdx
+++ b/docs/colors.mdx
@@ -17,11 +17,9 @@ import {
 
 Colors have their variations that may be used depending on the contextual need.
 
-### Usage 
+### Usage
 
 Instead of declaring a hexadecimal code like `#fff`, you may access all our colors via variables named in this pattern: `var(--color-space-100)` (which is white, if you look in the Space palette below). If you need the darkest shade of green, for example, use `var(--color-earth-1000)` and so forth.
-
-To be able to access the variables, though, you'll have to import the corresponding file directly on your file: `@import "@magnetis/astro/src/css/colors.css";`. This is necessary because our bundle converts all variables to hexadecimal code.
 
 If you do that, you should also convert the variables in your build to hexadecimal code, since CSS variables have [limited browser support](https://caniuse.com/#feat=css-variables). We recommend the plugin [PostCSS CSS Variables](https://github.com/MadLittleMods/postcss-css-variables).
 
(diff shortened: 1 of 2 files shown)


larimaza:
[Investigar impossibilidade de uso de variáveis do Astro em outros projetos](https://trello.com/c/1RwmE9lX/134-investigar-impossibilidade-de-uso-de-vari%C3%A1veis-do-astro-em-outros-projetos)

docs/colors.mdx
```diff
@@ -17,7 +17,13 @@ import {
 
 Colors have their variations that may be used depending on the contextual need.
 
-Usage: Instead of declaring a hexadecimal code like `#fff`, you may access all our colors via variables named in this pattern: `var(--color-space-100)` (which is white, if you look in the Space palette below). If you need the darkest shade of green, for example, use `var(--color-earth-1000)` and so forth.
+### Usage 
+
+Instead of declaring a hexadecimal code like `#fff`, you may access all our colors via variables named in this pattern: `var(--color-space-100)` (which is white, if you look in the Space palette below). If you need the darkest shade of green, for example, use `var(--color-earth-1000)` and so forth.
+
+To be able to access the variables, though, you'll have to import the corresponding file directly on your file: `@import "@magnetis/astro/src/css/colors.css";`. This is necessary because our bundle converts all variables to hexadecimal code.
```
luisfmsouza: Just curious: why do we prefer to import color variables locally instead of export them inside the Astro bundle?

elmarques: We use a plugin to replace CSS variables with fixed color values.

I thought it wasn't possible to build a hybrid version (where variables are retained, but colors are replaced).

Taking a look today at the documentation, I saw that it's possible.

@larimaza, so so so so sorry for not having seen this plugin config before. 😞 

> Options
preserve (default: false)
Allows you to preserve custom properties & var() usage in output.
Possible values:
false: Removes --var declarations and replaces var() with their resolved/computed values.
true: Keeps var() declarations in the output and has the computed value as a fallback declaration. Also keeps computed --var declarations.
'computed': Keeps computed --var declarations in the output. Handy to make them available to your JavaScript.

@larimaza I think you also prefer don't exclude vars from Astro bundle (to make a single import).

I added the plugin option and removed from docs the lines with the extra import instructions.

@luisfmsouza thank you for your comment. It generated this insight. 😊 

Changed in the commit 97fb660.

Reference: https://github.com/MadLittleMods/postcss-css-variables#options

larimaza: Wow, that's so much better 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏  Let's ship this!

PR #86, opened 2019-05-06
Create masked input components

larimaza:
# What

Create masked input components.

# Why

We have design specs for masked inputs and need to collect different types of masked data in forms.

# How

- Introduce [Input Mask](https://github.com/RobinHerbots/Inputmask) to the documentation (but not to the Astro bundle)

# Sample

_You can attach images to show us the result, if applicable._

### DIFF
--- a/docs/inputs.mdx
+++ b/docs/inputs.mdx
@@ -74,7 +74,7 @@ To create the disabled state, add the `a-input--disabled` modifier class on elem
 
 Inputs that have a specific value structure for certain kinds of data.
 
-Masked inputs don't need specific CSS classes other than the main one shown below. However, they do need a [mask plugin](https://github.com/text-mask/text-mask) and custom JS to work on your project. You may find the code for this working example in our GitHub repository. If you decide to use Inputmask like we did, please follow the example to get the same results.
+Masked inputs don't need specific CSS classes other than the main one shown below. However, they do need a [mask plugin](https://github.com/text-mask/text-mask) and custom JS to work on your project. You may find the code for this working example in our GitHub repository. If you decide to use Text Mask like we did, please follow the example to get the same results.
 
 <Playground className='two-columns'>
   <div className='a-input'>
--- a/docs/inputs.mdx
+++ b/docs/inputs.mdx
@@ -76,7 +76,7 @@ Inputs that have a specific value structure for certain kinds of data.
 
 Masked inputs don't need specific CSS classes other than the main one shown below. However, they do need a [mask plugin](https://github.com/text-mask/text-mask) and custom JS to work on your project. You may find the code for this working example in our GitHub repository. If you decide to use Inputmask like we did, please follow the example to get the same results.
 
-<Playground className='two-columns masked-inputs'>
+<Playground className='two-columns'>
   <div className='a-input'>
     <input
       id='inputdate'
@@ -168,7 +168,7 @@ The buttons will need JS to work on your project. You may find the code for this
 
 To create the disabled state, add the `a-input--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
 
-<Playground className='two-columns masked-inputs'>
+<Playground className='two-columns'>
   <div className='a-input a-input--control'>
     <input
       id='control1'
--- a/docs/inputs.mdx
+++ b/docs/inputs.mdx
@@ -84,7 +84,9 @@ The masked inputs need a [mask plugin](https://github.com/RobinHerbots/Inputmask
       aria-labelledby='labeldate'
       data-inputmask-showmaskonhover='false'
       data-inputmask-mask='99/99/9999'
-      data-inputmask-placeholder='DD/MM/YYYY'
+      data-inputmask-placeholder=''
+      data-inputmask-jitmasking='true'
+      placeholder='DD/MM/YYYY'
     />
     <label id='labeldate' htmlFor='inputdate'>
       Date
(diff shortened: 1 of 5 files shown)


docs/inputs.mdx
```diff
@@ -144,10 +199,16 @@ To create the disabled state, add the `a-input--disabled` modifier class on elem
     <label id='controllabel2' htmlFor='control2'>
       Control input (R$)
     </label>
-    <button className='a-btn a-btn--ghost-uranus a-btn--small a-btn--icon a-input--control__decrement' disabled >
+    <button
```
elmarques: Please don't consider these buttons. It was Prettier thing. <💅/>

docs/inputs.mdx
```diff
@@ -70,6 +70,67 @@ To create the disabled state, add the `a-input--disabled` modifier class on elem
   </div>
 </Playground>
 
+## masked inputs
+
+Inputs that have specific structure for certain kinds of data.
+
+The masked inputs need a [mask plugin](https://github.com/RobinHerbots/Inputmask) and custom JS to work on your project. You may find the code for this working example in our GitHub repository.
```
luisfmsouza: Intrusion here!

I think we will not be able to use this plugin inside Vue or React apps, because these libs have a different approach than `Inputmask`, which touches in the DOM directly.

My question is: if we probably will not use this plugin inside our apps, should we continuou recommending `Inputmask`?

elmarques: Great observation! 🦅 

We've double checked that Inputmask lib isn't well optimized for reactive apps.

We replace it with another lib certainly compatible and quite solid:

This new lib supports vanilla JS, Angular, Vue, React, etc.

We also provide code and documentation updates in commit 84e469f.

Staging:
https://deploy-preview-86--astro-staging.netlify.com/#/docs-inputs

luisfmsouza: Sorry to be me! 🙊  This lib is outdated and the last update was almost 1 year ago.

I'm not trying to block things here or just bother you with annoying comments. This first version of Astro is meant to be only for visual aspects of our apps. For me, masked inputs is more than that, because we are stepping into the JS world here.

I truly believe the responsibility to implement masked input are for the third-party (investor, magnetis_v2, etc) apps and not for Astro.

docs/inputs.mdx
```diff
@@ -122,13 +164,20 @@ These inputs are commonly used for money input data, in contexts that require in
 
 They carry outline [icon buttons](/docs-buttons#icon-buttons) and these will always add or subtract R$ 100 or R$ 1.000.
 
-The buttons will need JS to work on your project. You may find the code for this working example in our GitHub repository.
+The buttons will need JS to work on your project. You may find the code for this working example in our GitHub repository. We also used a plugin for the currency mask, as explained in the Masked Inputs section above.
 
 To create the disabled state, add the `a-input--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
 
-<Playground className='two-columns'>
+<Playground className='two-columns masked-inputs'>
```
brunobc182: Why do we need this class here? 🤔

docs/inputs.mdx
```diff
@@ -70,6 +70,48 @@ To create the disabled state, add the `a-input--disabled` modifier class on elem
...
+      Date
+    </label>
+  </div>
+  <div className='a-input'>
+    <input
+      id='inputmoney'
+      type='text'
+      placeholder='0,00'
+      aria-labelledby='labelmoney'
+    />
```
brunobc182: This input have decimal separation In our design specs, but i tested locally and doesn't work. 
Screen Shot 2019-05-07 at 14 57 48

It's seems to work on control input when i click the decrement or increment buttons, but if i type manually still the mask it doesn't place the decimal separator

Screen Shot 2019-05-07 at 15 08 07

docs/inputs.mdx
elmarques: Great catch! 👀 This class served the selector of the previous lib. It's no longer needed. Thanks! 🙏 

Removed in commit 3a81f1c.

docs/inputs.mdx
```diff
@@ -70,6 +70,48 @@ To create the disabled state, add the `a-input--disabled` modifier class on elem
   </div>
 </Playground>
 
+## masked inputs
+
+Inputs that have a specific value structure for certain kinds of data.
+
+Masked inputs don't need specific CSS classes other than the main one shown below. However, they do need a [mask plugin](https://github.com/text-mask/text-mask) and custom JS to work on your project. You may find the code for this working example in our GitHub repository. If you decide to use Inputmask like we did, please follow the example to get the same results.
```
larimaza:
```suggestion
Masked inputs don't need specific CSS classes other than the main one shown below. However, they do need a [mask plugin](https://github.com/text-mask/text-mask) and custom JS to work on your project. You may find the code for this working example in our GitHub repository. If you decide to use Text Mask like we did, please follow the example to get the same results.
```

docs/inputs.mdx
elmarques: @brunobc182, man, as far as I know, the usage of this mask needs to add the comma to have pennies.

The input control works because we did some ~gambiarras~ handling by JavaScript. 😬 

I think it may be very painful to adapt the plugin to have this behavior. 😱 

Do you think it would be too bad to proceed with the normal behavior of the lib?

brunobc182: That's fine, i don't have any problem with that. Just for your information, i implemented the currency mask a little different, you can see [here](https://github.com/magnetis/astro/blob/masked-input/docs/js/inputs.js#L145).

elmarques: Thanks for all, man! 🙏 I did some additional testing without success. Some almost worked. 😞 

I suggest merge this PR and open a tech debt (after 1.0 launch) to analyze the mask issues more carefully.

@larimaza Can I create a card for this in our backlog?

PR #88, opened 2019-05-07
Remove JS dependency from Astro

elmarques:
# What

This {spike} PR completely removes the JavaScript dependency of the Astro.

# Why

The first version of Astro should be made to deal only with the look and feel of the apps.

# How

* Remove all custom JS from code
* Remove all JS references from docs
* Adjust some styles to work without JS


larimaza:
We decided not to release this and keep JS in the project, after our discussion on Slack :)

PR #94, opened 2019-05-16
Add font smoothing to Typography (Chrome)

larimaza:
# What

Add font smoothing to our Typography to make the fonts render better in Chrome.

# Why

Fonts look bolder/heavier than they should on Chrome.
Couldn't find a property that worked on Firefox yet, so this PR only addresses Chrome.

# How

Add `-webkit-font-smoothing: antialised` to typography CSS.
It appears that our prefixer doesn't work with this property, so we had to add `-webkit-`.

### DIFF
--- a/src/css/typography.css
+++ b/src/css/typography.css
@@ -9,6 +9,7 @@
 .a-text--small {
   font: var(--font-primary);
   color: var(--font-color);
+  -webkit-font-smoothing: antialiased;
 }
 
 .a-text--secondary-large,
@@ -16,6 +17,7 @@
 .a-text--secondary-small {
   font: var(--font-secondary);
   color: var(--font-color);
+  -webkit-font-smoothing: antialiased;
 }
 
 /* Display */
--- a/src/css/typography.css
+++ b/src/css/typography.css
@@ -10,6 +10,7 @@
   font: var(--font-primary);
   color: var(--font-color);
   -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
 }
 
 .a-text--secondary-large,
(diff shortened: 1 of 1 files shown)


src/css/typography.css
```diff
@@ -9,13 +9,15 @@
...
   color: var(--font-color);
+  -webkit-font-smoothing: antialiased;
 }
 
 .a-text--secondary-large,
 .a-text--secondary-medium,
 .a-text--secondary-small {
   font: var(--font-secondary);
   color: var(--font-color);
+  -webkit-font-smoothing: antialiased;
```
AgtLucas: Let's also provide the Firefox alternative:

```css
-moz-osx-font-smoothing: grayscale;
```

What do you think?

src/css/typography.css
```diff
@@ -9,13 +9,15 @@
 .a-text--small {
   font: var(--font-primary);
   color: var(--font-color);
+  -webkit-font-smoothing: antialiased;
```
kuroski: I think in firefox you can use the:
`-moz-osx-font-smoothing: grayscale;` property

https://medium.com/@MateMarschalko/improving-font-rendering-with-css-3383fc358cbc

src/css/typography.css
larimaza: I tested this, but it didn't seem to make any difference at all in my computer 🤔 But we can include that just to be safe, I guess.

src/css/typography.css
larimaza: I tested this, but it didn't seem to make any difference at all in my computer 🤔 But we can include that just to be safe, I guess.

PR #96, opened 2019-05-20
Extract font-sizes values into variables

AgtLucas:
# What

To improve the development of new components, as we discussed, I'm extracting common used values/tokens into variables. In this case, **font sizes**.

# Why

We have a lot of values/tokens hard coded throughout our codebase, extracting them into variables makes it easy to update values across the codebase if needed.


src/css/fonts.css
```diff
@@ -2,4 +2,12 @@
   --font-primary: normal normal 600 normal 16px/1.5 Poppins, sans-serif;
   --font-secondary: normal normal 400 normal 16px/1.5 Lato, sans-serif;
   --font-color: var(--color-moon-900);
+  --font-size-small: 12px;
```
kuroski: I like the idea of standardizing font spacing and sizes with variables.
The hard part is naming things :smile: 

This post is not about font-size variables naming conventions, but you can use the same "idea" behind that:
http://bradfrost.com/blog/post/7-habits-of-highly-effective-media-queries/#major-minor

Instead of having something like:

``` css
--font-size-small: 12px;
.....
--font-size-xlarge: 32px;
--font-size-xxlarge: 40px; // hmm ok
--font-size-xxxlarge: 48px; // here thinks start to get messy
--font-size-huge: 64px;
```

that we may have in the future something like:

``` css
--font-size-xxxxlarge: 50px; // we just need
--font-size-hhuge: 70px; // I don't know how to call this thing >_<
--font-size-colossal: 100px;
```

We may just have something like:
(I'm using naming conventions strip from [utility first css frameworks](https://tailwindcss.com/docs/font-size) since they think in better naming from attributes)

``` css
--font-size-sm: 12px;
--font-size-base: 16px;
--font-size-lg: 24px;
--font-size-xl: 32px;
--font-size-2xl: 40px;
--font-size-3xl: 48px;
--font-size-4xl: 64px;
```

That you can then add new variables without naming problems:

``` css
--font-size-2sm: 14px;
--font-size-2lg: 28px;
```

And you can use different syntax for the variables:
- text-sm / text-lg / text-xl / text-xl-2
- font-size-sm / font-size-lg / font-size-xl-2
- text-small / text-large / text-xlarge / text-xlarge-2

Or you can use the naming of the astro documentation that the designers used:

``` css
--text-small-poppins: --px;
--text-medium-poppins: --px;
--text-large-poppins: --px;
--title-small-poppins: --px;
--title-medium-poppins: --px;
--title-large-poppins: --px;
....
```

what do you think?

AgtLucas: @kuroski Yep, naming things is really hard 🙈

I like the idea this one:

```css
--font-size-sm: 12px;
--font-size-base: 16px;
--font-size-lg: 24px;
--font-size-xl: 32px;
--font-size-2xl: 40px;
--font-size-3xl: 48px;
--font-size-4xl: 64px;
```

@luisfmsouza suggested using something like this:

```css
--font-size-base: 8px;

--font-size-2base: calc(--font-size-base * 2);
--font-size-3base: calc(--font-size-base * 3);
/* and so on */
```

I like both approaches! We need to decide for sure about naming.

AgtLucas: @kuroski I don't like this one though:

```css
--text-small-poppins: --px;
--text-medium-poppins: --px;
--text-large-poppins: --px;
--title-small-poppins: --px;
--title-medium-poppins: --px;
--title-large-poppins: --px;
```

- If we change our fonts, we need to change the variables names;
- We might ending up with duplicated values in different variables;

luisfmsouza: I'd vote for something more pragmatic avoiding to mention `small`, `large`, etc which is an arbitrary definition.

Our font definition is based on `8px` scale, so we could have a `--font-size-base` variable with the value `8px`.

Following this, we can use the CSS `calc()` to create the other sizes:

```css
  --font-size-2base: calc(--font-size-base * 2);
  --font-size-3base: calc(--font-size-base * 3);
  --font-size-4base: calc(--font-size-base * 4);
  --font-size-5base: calc(--font-size-base * 5);
  --font-size-6base: calc(--font-size-base * 6);
  --font-size-8base: calc(--font-size-base * 8);
```

If we want to have a default variable, to help coders to understand what's the meaning of some variable, we can create aliases:

```css
--font-size-default: --font-size-2base;
--font-size-huge: --font-size-8base;
```

The problem would be the `12px` size name. This suggestion is not terrible, but lacks on clarity:

```css
  --font-size-1-5base: calc(--font-size-base * 1.5);
```

WYT?

AgtLucas: I like the base one as well!

brunobc182: I really enjoy the `--font-size-base` calculation, it's easy and simple to understand. I think we can document this to help new contributors.

Nice work, btw!

larimaza: Definitely agree with the last comment from @luisfmsouza, minus the aliases because we should keep things simple.
Anyway, this should be something that Astro users won't touch a lot, because they should use the Typography classes. Contributors can read the documentation (which we should update in this very PR, I can help with that) to understand the pattern :)

AgtLucas: Cool, so we are going with the base approach!

src/css/fonts.css
```diff
@@ -2,4 +2,12 @@
   --font-primary: normal normal 600 normal 16px/1.5 Poppins, sans-serif;
```
larimaza: Would it work if we replaced this "16px" too with the `--font-size-2base` variable? 🤔

luisfmsouza: think this would work @larimaza 

```suggestion
  --font-primary: normal normal 600 normal var(--font-size-2base)/1.5 Poppins, sans-serif;
```

AgtLucas: @larimaza Ah, nice catch! Will do that ASAP!

AgtLucas: @larimaza @luisfmsouza Done!

AgtLucas:
@luisfmsouza Build is passing now, 'twas failing because of the variables order declaration. Fixed here: https://github.com/magnetis/astro/pull/96/commits/5f529c122bde35bfc34b5460e7049c2b34393756

PR #98, opened 2019-05-22
Add media query for showing inputs in two columns

kuroski:
# What

This commits changes the default behavior of showing the inputs in the documentation.

# Why

The inputs in the documentation are always showed with two columns and when the screen is too small they look very squishy.

# How

Now they only have two columns if the screen is large enough using the class `md:two-columns` in the playground tag.

# Sample

sample

### DIFF
--- a/doczrc.js
+++ b/doczrc.js
@@ -251,9 +251,11 @@ export default {
         '& > .a-slider': {
           marginBottom: 22
         },
-        '&.two-columns': {
-          display: 'inline-grid',
-          gridTemplateColumns: 'auto auto'
+        '@media (min-width: 780px)': {
+          '&.md\\:two-columns': {
+            display: 'grid',
+            gridTemplateColumns: 'auto auto'
+          }
         },
         '&.gradient-bg': {
           backgroundImage: 'var(--gradient-andromeda)'
--- a/docs/inputs.mdx
+++ b/docs/inputs.mdx
@@ -21,7 +21,7 @@ The floating label will need JS to work on your project. You may find the code f
 
 To create the disabled state, add the `a-input--disabled` modifier class on element root and `disabled` attribute on the `<input>` tag.
 
-<Playground className='two-columns'>
+<Playground className='md:two-columns'>
   <div className='a-input'>
     <input id='input1' type='text' aria-labelledby='label1' />
     <label id='label1' htmlFor='input1'>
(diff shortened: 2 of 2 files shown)


kuroski:
[ch7522]

kuroski:
[ch7522]

doczrc.js
```diff
@@ -251,9 +251,11 @@ export default {
         '& > .a-slider': {
           marginBottom: 22
         },
-        '&.two-columns': {
-          display: 'inline-grid',
-          gridTemplateColumns: 'auto auto'
+        '@media (min-width: 780px)': {
+          '&.md\\:two-columns': {
```
larimaza: I've never seen CSS classes with `:`, nor this `\\` thing 🤔

kuroski: \\ is just to escape the `:`

Normally we only use `\`, but because it's compiled css through js we need to scape two times :sweat_smile: 

And we can use `:` to indicate medias (:exclamation: and nothing else)

So we can stick with the BEM or X convention

```
:root {
  --bp-small: 400px;
  --bp-small-2: 500px;
  --bp-medium: 800px;
}

.my-class {}

@media (min-width: 400px) {
  .sm\:my-class {} // to apply "sm:my-class" only to screens > 400px
}

@media (min-width: 500px) {
  .sm2\:my-class {} // to apply "sm2:my-class" only to screens > 500px
}
```

larimaza: Ohh okay :D so cool, thanks for explaining!

elmarques: 🤯 WOW! 🤯

I didn't know that was possible. 👏 

I found this to complement my learning:
https://pineco.de/css-quick-tip-the-valid-characters-in-a-custom-css-selector/

It seems that [Tailwind CSS](https://tailwindcss.com) uses this approach.

Stay tuned, the OCD monster will attack! 🤣

luisfmsouza: Just curious: can we achieve the same result avoiding the `md:` @kuroski ?

kuroski: @elmarques yes uhauhahuahu some other tools tend to use that for medias

@luisfmsouza yeah we could:
- just put a media query in the original `two-columns` class
- call them something like: `two-columns--md` / `two-columns-md` or something like that (there are several ways we can name this thing T_T

luisfmsouza: I'd vote for something we don't need to escape in the CSS side. Just imagine if someone, which is not in this conversation, tries to find this class, would be easy for this person? This is almost the same problem with SASS-like classes. What you think?

PR #99, opened 2019-05-22
Update Readme with usage info

larimaza:
# What

Update Readme to include detailed usage info.

# Why

Some projects inside Magnetis were using Astro in a different way than intended, so we're documenting this more clearly.

# How

Update README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,17 +85,35 @@ In other words, avoid this:
 
 ```css
 .paragraph {
-  font-family: var(--font-primary); 
-  font-weight: 600; 
-  font-size: 24px; 
-  line-height: 1.5; 
-  color: var(--color-moon-900); 
+  font-family: var(--font-primary);
+  font-weight: 600;
+  font-size: 24px;
+  line-height: 1.5;
+  color: var(--color-moon-900);
 }
 ```
+```html
+<p class="paragraph">Don't do this</p>
+```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,6 +79,24 @@ Here's a quick example. If you'd like to create a Display text styled in Astro t
...
+  font-weight: 600; 
+  font-size: 24px; 
+  line-height: 1.5; 
+  color: var(--color-moon-900); 
+}
+```
+
+In the example above, all these properties could be replaced with using the `a-text--large` class in the `paragraph` element. 
+
+In case you absolutely have to customize one or more properties in an Astro component, you should still use the corresponding Astro class that's closer to your goal and override it with your custom properties.
```
luisfmsouza: Think we can explain the reasons, but for me some examples would do a great work here. What you think about something like:

### Bad

```css
.paragraph {
  font-family: var(--font-primary); 
  font-weight: 600; 
  font-size: 24px; 
  line-height: 1.5; 
  color: var(--color-moon-900); 
}
```

```html
<p class="paragraph">this is a BAD paragraph!</p>
```

### Good

```css
.paragraph {
  color: var(--color-moon-700); /* a color we want to overwrite */
}
```

```html
<p class="a-text--large paragraph">this is a GOOD paragraph!</p>
```

larimaza: Done :) 
563d4d5

PR #100, opened 2019-05-22
Update Iconography with new icons

larimaza:
# What

Update Iconography with three new icons: chart-up, comparison and flag.

# Why

We got new icons in Astro 🎉 

# How

- Add new SVG files to the assets folder;
- Add them in our CSS icon classes;
- Add them to the Iconography docs.

# Sample

Screen Shot 2019-05-22 at 16 12 42Screen Shot 2019-05-22 at 16 12 51Screen Shot 2019-05-22 at 16 12 58


assets/icons/chart-up.svg
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kuroski: Did you compress these icons?
https://jakearchibald.github.io/svgomg/

larimaza: There's no need because we have a dependency that compresses all of them in the bundle :)

PR #113, opened 2019-06-05
Fix control input missing button

larimaza:
# What

Fix the mystery of the missing Control Input decremental button.

# Why

We realized that, in production, one of the Control Input buttons was missing (the decremental one), replaced with a color block.
In further investigation, the SVG is still in the assets folder, but the class is not in the `icons.css` file. WTF?

# How

Create the class to fix the issue.

# Sample

Screen Shot 2019-06-05 at 17 44 31

### DIFF
--- a/src/css/icons.css
+++ b/src/css/icons.css
@@ -256,6 +256,10 @@
   mask-image: url(../../assets/icons/light.svg);
 }
 
+.a-icon__input-dash {
+  mask-image: url(../../assets/icons/input-dash.svg);
+}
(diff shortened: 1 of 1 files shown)


larimaza:
When you investigate a WTF moment and realize that the person who messed up was yourself:

https://github.com/magnetis/astro/commit/cbee45e9df9002f39cc5953e2be0d9253070710d#diff-015728fd7b45a495c786a441a88cc3bc

PR #115, opened 2019-06-05
Takes off the need to manually import Astro fonts

LopesAlysson:
# What

Takes off the need to manually import Astro fonts in the HTML of the projects where it will be used.

# Why

To make easier to other developers use this project, since they will only have to **yarn add** and nothing more.

# How

Removes docz.html `<link>` tags to the *Lato* and *Poppins* fonts and import them via `@import` in index.css

# Sample

None


LopesAlysson:
> It would be correct for us to force the use of the fonts?
> 
> Because let's say that the user already have these fonts in the project, we would simply be forcing him to remove them. right?

@kuroski 
I think it is a question of philosophy to follow one or another approach. In most of the cases I believe that devs will not have this fonts already imported. But it will be really bad for the ones who have alredy imported Astro fonts somewhere else.

So, in conclusion is a decision to make to deliver a *ready to use* package or add some additional configuration in order to guarantee the best performance and personalization.

LopesAlysson:
> There was an old-school recommendation that we should avoid the use of CSS imports for performance reasons.
> 
> https://www.stevesouders.com/blog/2009/04/09/dont-use-import/
> https://gtmetrix.com/avoid-css-import.html
> 
> Does this concern still make sense in ours _fast-internet-and-modern-browsers_ days?

Hi, @elmarques !

Yes, it does still make sense to be concerned about `@imports` because they're a blocking calls, oposing to `<link>` imports, that can run in pararell. And even being only two `@import` calls, it could impact in browser performance on the first render.

Actually, I'm starting to have second thoughts if this feature we would be introducing really pays off the downside of it.

kuroski:
> > It would be correct for us to force the use of the fonts?
> > Because let's say that the user already have these fonts in the project, we would simply be forcing him to remove them. right?
> 
> Hey @kuroski ,
> I think it is a question of philosophy to follow one or another approach. In most of the cases I believe that devs will not have this fonts already imported. But it will be really bad for the ones who have alredy imported Astro fonts somewhere else.
> 
> So, in conclusion is a decision to make to deliver a _ready to use_ package or add some additional configuration in order to guarantee the best performance and personalization.

Thanks a lot, you're right, I'll be approving this =D

kuroski:
> > There was an old-school recommendation that we should avoid the use of CSS imports for performance reasons.
> > https://www.stevesouders.com/blog/2009/04/09/dont-use-import/
> > https://gtmetrix.com/avoid-css-import.html
> > Does this concern still make sense in ours _fast-internet-and-modern-browsers_ days?
> 
> Hi, @elmarques !
> 
> Yes, it does still make sense to be concerned about `@imports` because they're a blocking calls, oposing to `<link>` imports, that can run in pararell. And even being only two `@import` calls, it could impact in browser performance on the first render.
> 
> Actually, I'm starting to have second thoughts if this feature we would be introducing really pays off the downside of it.

The `@import` it's not an issue, since we use postCSS
So at the end, the final user will always use a compiled version of `Astro`.

And if anyone use webpack and `import `@magnetis/astro/index.css` it wont be a problem either since it will be compiled to a single css file at the end.

elmarques:
> The @import it's not an issue, since we use postCSS
So at the end, the final user will always use a compiled version of Astro.

@kuroski, I think to make the postCSS embed the imported file content it's necessary a plugin:

https://github.com/postcss/postcss-import

Otherwise, it keeps the `@import` unchanged.

Captura de Tela 2019-06-24 às 14 15 45

Maybe `postcss-import` plugin might not work with remote URLs.

I found another postCSS plugin that seems to do the job:

https://github.com/jonathantneal/postcss-font-magician

Maybe this way we can embed the fonts without `@import` performance issues.

What do you think, @LopesAlysson?

LopesAlysson:
> What do you think, @LopesAlysson?

I think it's a wonderful solution! That way we can avoid the caveats of this PR.
I'll try that and let you all know.

LopesAlysson:
Hi, guys, so I've found an amazing source about font optimization and the bad news is that there's no scape from blocking requests while downloading fonts but there are several points we can be careful in order to improve the user experience during render time. Here's the source:  https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization

I think it would be really nice to use Font Magician because it already has all this optimizations inside.

LopesAlysson:
Experimenting with Font Magician, I realized that the optimization he was doing was far inferior to the one done by Google Fonts on the import url. So, if the main concern is about the @import, maybe we could only download the file and use it, since [this url](https://fonts.googleapis.com/css?family=Lato:400,700|Poppins:500,600,700&display=swap) only points to a CSS file with @font-faces declarations.

I've done it here: 0ae1828debc8b2c7c8a28a39a8530b7765708c9b

LopesAlysson:
[ch7506]

docs/components/GetStartedCards.jsx
```diff
@@ -14,15 +14,13 @@ export default () => (
...
-            <a href='https://astro.magnetis.com.br/typography'>
-              Typography page
-            </a>{' '}
-            and load them in your project.
-          </p>
+          <p>Or via CDN:</p>
+        </li>
+        <li className='code-snippet'>
+          <pre>
+          {'<'}link rel='stylesheet' href='https://unpkg.com/@magnetis/astro/dist/astro.css' />
```
AgtLucas: @LopesAlysson Is this right?

AgtLucas: This:

```js
{'<'}
```

larimaza: It's a little guy with a pointy nose and no mouth 😄

AgtLucas: @larimaza I thought it was a 🐦 😂

LopesAlysson: Yes, it is!

If I don't write it that way, docz will try top render as a HTML tag, therefore the need to parse as a string. :)

PR #125, opened 2019-07-03
Update slider thumb size and add full fill

larimaza:
# What

Update slider thumb size to improve touch area in mobile devices;
Add the full fill variant using our gradients.

# Why

Current thumb size is too small to control in a touchscreen.

# How

Update CSS and docs to reflect the changes in all browsers.

# Sample

**Before** | **After**
-|-
Screen Shot 2019-07-03 at 15 19 41|Screen Shot 2019-07-03 at 15 22 30


kuroski:
[ch11054]

larimaza:
> Nice work =D
> 
> Just one question, the `autoprefixer` wont get those css selectors variants? (`-moz/-ms/-webkit`)?

@kuroski Nope :( The autoprefixer only works in properties, not in weird pseudo selectors.

PR #133, opened 2019-07-17
Remove ghost buttons from the docs

larimaza:
# What

Deprecate the primary text version of ghost buttons in the docs.

# Why

We are deprecating the primary text use of ghost buttons because nothing visually indicates that they're a clickable element, so they're bad for accessibility and usability.

# How

As we still support ghost iconlabel buttons and ghost icon buttons, we still need to leave all the CSS untouched. So, this PR only removes the Ghost Button sections from the documentation.

* This task _was_ going to be a part of version 2.0.0, but  it will be merged to master instead as there's no breaking changes after all.

### DIFF
--- a/docs/buttons.mdx
+++ b/docs/buttons.mdx
@@ -55,30 +55,6 @@ Mouseover and click to view `:hover` and `:active` states.
   </button>
 </Playground>
 
-## ghost buttons
-
-Ghost buttons are visually similar to inline links, but with the size and behavior of regular buttons. They can be used for secondary actions or microinteractions.
-
-Aside from using the ghost button CSS class, you also need to specify size and color.
-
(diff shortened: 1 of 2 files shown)


larimaza:
[ch11053]

larimaza:
> 💥
> 
> Just one question, are you sure that no one here is using those classes?

If they are, nothing is going to break because we didn't remove anything from the CSS, just the docs :)

larimaza:
> 💥
> 
> Just one question, we're removing from docs but still have support? >_<

Theoretically we won't support them, but it still makes sense to keep the same ghost class structure for iconlabel/icon buttons that still use them, as there are still two variants using that ghost base (and there could be more in the future)

PR #134, opened 2019-07-17
Fix bugs related to input invalid and validated states

VictorHeringer:
# What

This __PR__ fix some bugs related to inputs when it has the invalid or validated states. 

# Why

* Invalid state for empty inputs was causing the placeholder to have a color that makes the placeholder looks like its some value instead of an actually placeholder.
* When `.a-input--validated` or `.a-input--invalid` was used at control inputs, the icon related to this two classes was overleaping with its buttons.

# How

* Remove icon from invalid and validated state for control input
* Add examples at docs for invalid and validated control input
* Change `.a-input--validated > label ` (it was the only one with different color)
* Change `.a-input--invalid > label` to have default placeholder color instead of invalid
# Sample

Invalid and validated control input example
Captura de Tela 2019-07-17 às 17 37 07

Keeping the `--color-moon-400` color for placeholders even when its invalid
ezgif com-video-to-gif

### DIFF
--- a/docs/inputs.mdx
+++ b/docs/inputs.mdx
@@ -212,6 +212,45 @@ To create the disabled state, add the `a-input--disabled` modifier class on elem
       <i className='a-icon a-icon__input-plus a-icon--size-small' />
     </button>
   </div>
+  <div className='a-input a-input--control a-input--validated'>
+    <input
+      id='control3'
+      type='text'
+      data-step='100'
+      defaultValue='0,00'
+      placeholder='0,00'
+      aria-labelledby='controllabel3'
+    />
+    <label id='controllabel3' htmlFor='control3'>
+      Control input validated (R$)
+    </label>
+    <button className='a-btn a-btn--ghost-uranus a-btn--small a-btn--icon a-input--control__decrement'>
+      <i className='a-icon a-icon__input-dash a-icon--size-small' />
+    </button>
+    <button className='a-btn a-btn--ghost-uranus a-btn--small a-btn--icon a-input--control__increment'>
+      <i className='a-icon a-icon__input-plus a-icon--size-small' />
+    </button>
+  </div>
+  <div className='a-input a-input--control a-input--invalid'>
+    <input
+      id='control4'
+      type='text'
+      data-step='100'
+      defaultValue='0,00'
+      placeholder='0,00'
+      aria-labelledby='controllabel4'
+    />
+    <label id='controllabel4' htmlFor='control4'>
+      Control input invalid (R$)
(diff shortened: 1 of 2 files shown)


docs/inputs.mdx
```diff
@@ -212,6 +212,45 @@ To create the disabled state, add the `a-input--disabled` modifier class on elem
       <i className='a-icon a-icon__input-plus a-icon--size-small' />
     </button>
   </div>
+  <div className='a-input a-input--control a-input--validated'>
+    <input
+      id='control1'
```
larimaza: We should change these ids and respective labels to make them unique, seems like we have lots of 'control1'...

src/css/inputs.css
```diff
@@ -141,14 +141,14 @@
 }
 
 .a-input--validated > label {
-  color: var(--color-moon-300);
+  color: var(--color-moon-400);
```
larimaza: Why did we change this color from 300 to 400? Just curious

src/css/inputs.css
```diff
@@ -141,14 +141,14 @@
...
 }
 
 .a-input--validated > input:focus + label {
   color: var(--color-earth-400);
 }
 
-.a-input--validated::after {
+.a-input--validated:not(.a-input--control)::after {
```
larimaza: 👏

src/css/inputs.css
VictorHeringer: For some reason the `a-input--validated` was having `--color-moon-300` as placeholder color, I notice this slight diference assume that all placeholder should have same color to keep consistence and change it. But now I actually don't know if this diferente color is proposital or not 🤔. Any thoughts on this?

At design specs looks like the placeholder/floating label has two different tones of gray (`--color-moon-300` and `--color-moon-400`) which could be the cause of this whole confusion. 

This is how it is at [astro](https://astro.magnetis.com.br/#/docs-inputs) today
Captura de Tela 2019-07-18 às 14 05 30

docs/inputs.mdx
VictorHeringer: Thanks, the ctr+c ctr+v got me ahahaha!

VictorHeringer: Fixed 9a460672942e5f5a33990652430213728dabddb7 😬

src/css/inputs.css
larimaza: Oh, I see! Yes, it was probably a design inconsistency, you were right changing to 400 to match the rest :)

larimaza:
ch12519

PR #140, opened 2019-07-31
[BREAKING - DON'T MERGE UNTIL 2.0] Rename Astro icons to generic names

larimaza:
# What

Rename a few Astro icons to more generic names.

# Why

Some names were very specific of the Magnetis context.
We decided it's better to keep these names more generic for an open source mindset.

# Rename list:

wallet -> chart
referral -> invite
trades -> time
redemption -> down
document -> sheet
page -> document


larimaza:
[ch11052]

assets/icons/sheet.svg
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
brunobc182: Could we optimize this svg with [optimage](https://getoptimage.com/), for example?

larimaza: We don't need to pre-optimize them because we have an SVG optimizer dependency 👯

PR #141, opened 2019-07-31
FIx minor layout breaks introduced by PR #137

larimaza:
# Ticket
[ch15144]

# What

Fix minor layout problems introduced by PR #137 

# Why

The following components look funny:

- checkboxes contents are not vertically aligned;
- the body vertical scrollbar is sometimes not visible;
- page header title is not responsive anymore;
- the ghost input playground background has a padding that cannot be selectively removed;
- line height is smaller in the body selector, which alters layout spacing all over the docs;
- menu width is larger than the original.

# How

Adjust CSS styles here and there.

# Note 

After this, we should test the checkbox components out of the Docz context to see if they're still ok.

### DIFF
--- a/docs/theme/index.js
+++ b/docs/theme/index.js
@@ -15,6 +15,7 @@ const selectors = {
   mobileMenu: '& .Sidebar__Wrapper-dMteFe',
   playgroundWrapper: playgroundSelector,
   playgroundChildrens: `${playgroundSelector} > div`,
+  headerTitle: '& .H1-ecIwba',
 
   /* Docz default elements (hidden) */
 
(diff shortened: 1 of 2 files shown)


docs/theme/index.js
```diff
@@ -25,12 +26,15 @@ const selectors = {
   menuBlueTop: '& .Logo__Wrapper-bzwKYh:before',
   menuItemsBorder: '& .MenuLink__Wrapper-bIIltX:after',
   menuSubitemsBorder: '& .MenuLink__Wrapper-bIIltX.gUhHxl:after',
-  headingIcon: '& .heading--Icon'
+  headingIcon: '& .heading--Icon',
```
brunobc182: I think we do not need this coma here. 💅

larimaza: Good catch!

PR #142, opened 2019-07-31
Improve error state for control input

VictorHeringer:
# What

This __PR__ improves the invalid state for control inputs, remove docs for valid state and fix a bug  related to ghost input valid and invalid state.

[ch15104](https://app.clubhouse.io/magnetis/story/15104/atualizar-estado-inv%C3%A1lido-do-control-input)

# Why

* Design specs for control input component has changed
* To remove a bug introduced previously

# How

* Hide the validation icons only at `a-input--control`
* Add colors to control input buttons on error state
* Remove valid state for input control from docs

# Sample

Bug fix for ghost inputs

|  Before | After  |
|---|---|
| Captura de Tela 2019-07-31 às 16 29 06  |  Captura de Tela 2019-07-31 às 16 28 35 |

Improve invalid control input

ezgif com-video-to-gif (1)

### DIFF
--- a/src/css/inputs.css
+++ b/src/css/inputs.css
@@ -69,8 +69,8 @@
   color: var(--color-space-100);
 }
 
-.a-input--ghost.a-input--validated::after,
-.a-input--ghost.a-input--invalid::after {
+.a-input--ghost.a-input--validated:not(.a-input--control)::after,
+.a-input--ghost.a-input--invalid:not(.a-input--control)::after {
   background-color: var(--color-space-100);
 }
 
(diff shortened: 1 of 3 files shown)


src/css/inputs.css
```diff
@@ -148,7 +165,7 @@
   color: var(--color-earth-400);
 }
 
-.a-input--validated:not(.a-input--control)::after {
```
larimaza: I'm not sure I agree with removing this logic to replace it with `.a-input--control.a-input--invalid::after {display: none;}`. 🤔 If both work the same way, I would think it's better to not apply a style to control inputs instead of applying it _and_ removing it... Is there a particular reason for this change?

src/css/inputs.css
```diff
@@ -166,11 +183,12 @@
   border: 1px solid var(--color-mars-500);
 }
 
-.a-input--invalid > input:focus + label {
+.a-input--invalid > input:focus + label,
+.a-input--invalid > .a-label--invalid {
```
larimaza: I'm thinking if there's a way we could solve this without the extra `a-label--invalid` class 🤔

larimaza: Maybe there's not 😅

src/css/inputs.css
VictorHeringer: Completely agree with you. I ran into a precedence problem but din't notice that I should've used the same precedence for ghosts inputs to solve the issue. Thanks for pointing this 😁  

Fixed f8744eefad531b004cad1506de7172c77e459e0e

PR #145, opened 2019-08-02
Fix variable bugs in checkboxes, toggles and radio buttons

larimaza:
# What

When building all our CSS to the final `astro.css` file, a few size variables are `undefined` and cause checkboxes, toggles and radio buttons to render incorrectly when importing CSS from that compiled file.
This PR fixes that issue.

# Why

The variables are defined inside CSS selectors in the checkboxes, toggles and radio buttons respective CSS files. That is what caused the issue.

# How

Remove the variables from the CSS selectors to define them inside `:root` instead and rebuild the files.


larimaza:
> Nice work.
> 
> Just to make sure, this will change only the checkbox variables, right?
> 
> After answer this 🚢

Yes, these variables only affect their respective scoped CSS files (checkboxes, toggles and radio buttons)

larimaza:
ch15353

larimaza:
[ch15353]

PR #150, opened 2019-08-12
Remove font declaration from disabled input

larimaza:
# What

Remove font declaration from disabled input label selector.

# Why

Since disabled is just a state, it's not necessary to set the base font again. Furthermore, this overrides other styles undesirably, such as when an input has a floating label *and* is disabled.

# How

Remove the font shorthand from the disabled input label selector.


larimaza:
[ch16024]

PR #153, opened 2019-08-14
Add step on deploy.js to create 404.html file

LopesAlysson:
# What

Fixes sublink problem that ended up returning a 404 error page on github pages.

# Why

It was not possible to go directly to a sublink without first accessing the home page.

# How

Add a new step on `node deploy.js` to copy `index.html` file as `404.html` allowing a redirect to sublinks starting with a hash (#).

### DIFF
--- a/deploy.js
+++ b/deploy.js
@@ -3,12 +3,17 @@ const gh = require('gh-pages');
 
 const DIST_PATH = '.docz/dist';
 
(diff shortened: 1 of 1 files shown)


LopesAlysson:
[ch16809]

larimaza:
SACRED MONSTER

PR #154, opened 2019-08-14
Create icon link buttons and add them to the documentation

larimaza:
# What

Create new icon link buttons and add them to the documentation.

# Why

This new type of button came with recent design specs - more specifically, the new home.

# How

- Create the styles in buttons.css;
- Add examples to the documentation;
- Change the background of `playground` to a different shade by design request.

# Sample

Screen Shot 2019-08-14 at 17 58 19


larimaza:
[ch16765]

PR #165, opened 2019-09-11
Create a section for file preview component

victorheringer:
# What

This __PR__ adds a new component to display previews for uploaded files. It's is important to notice the it is not yet related to an input file component.

# Why

We need to standardize the way that uploaded files are been previewed at different places.

# How

- Create image for folders and add Magnetis logo
- Create a section to display the examples at docs
- Create style for preview files from input

# Sample

Captura de Tela 2019-09-11 às 17 30 27

# Reference

[ch18872](https://app.clubhouse.io/magnetis/story/18872/criar-input-do-tipo-upload-em-inputs)


docs/inputs.mdx
```diff
@@ -158,6 +159,31 @@ This input style does not have a disabled state.
...
+## uploaded files preview
+
+Used to display a preview for uploaded files.
+
+If you don't wanna handle the displayed text size by yourself like in the first example, we offer a `a-input--file__text-truncate` class that can be
+used to truncate the text size. Just put it like `<p class="a-text--secondary-small a-input--file__text-truncate">` and it
+will use `text-overflow: ellipsis` to do the trick.
+
+<Playground>
+  <div className="a-input--file">
```
kuroski: This seams not following the BEM conventions

`a-input--file` is a modifier, so to use that you must use something like `a-input a-input--file`

In this case, this seams to be a block, so the name should be something like: `a-input-file`

docs/inputs.mdx
```diff
@@ -158,6 +159,31 @@ This input style does not have a disabled state.
...
+    <div className="a-input--file__preview-container">
```
kuroski: This naming convention is invalid, you can't declare an element of a modifier.

```
// OK
.a-input
.a-input--large
.a-input__label
.a-input__label--small

// Not so OK
.a-input--large__label
.a-input--large__label--small 
```

In this case, this element should be something like `a-input-file__preview`

docs/inputs.mdx
victorheringer: I guess the main issue was using this `--file` modifier 😅 if a correct this to be a block like `a-input-file` will solve both cases right? 0d5e02f93904d0c30a86630dc8c0b5a213ac0eab

larimaza: Hmm I don't think this is a block! This PR is for the "file input" **type of input**, similar to how "control input" is a type of input. So, if the class `input--control` is valid, so is `input--file` 😄

docs/inputs.mdx
larimaza: If we keep the `input--file` modifier naming, we'll have to fix this another way 😬

docs/inputs.mdx
larimaza: Can we review this before merging this PR? I'm a little confused with the class naming as it is now...

src/css/inputs.css
```diff
@@ -114,6 +114,76 @@
   background-color: var(--color-mars-400);
 }
 
+/* File inputs */
+
+.a-input-file .a-input-file__preview-container {
```
luisfmsouza: Just a small suggestion

```suggestion
.a-input-file .a-input-file__container {
```

docs/inputs.mdx
```diff
@@ -158,6 +159,29 @@ This input style does not have a disabled state.
...
+## uploaded files preview
+
+Used to display a preview for uploaded files.
+
+If you don't wanna handle the displayed text size by yourself like in the first example, we offer a `a-input--file__text-truncate` class that can be
+used to truncate the text size. Just put it like `<p class="a-text--secondary-small a-input--file__text-truncate">` and it
+will use `text-overflow: ellipsis` to do the trick.
+
+<Playground>
+  <div className="a-input-file">
```
lucasmazza: I might be late to the party, but as the PR states:

> It's is important to notice the it is not yet related to an input file component.

Maybe the component shouldn't have `input` on it? `a-file-preview` or `a-upload-preview` comes to mind as an alternative.

larimaza: That makes perfect sense, although someday that component *will* be a block inside an upload input component (in Astro React). 
Since we're suffering from a severe case of analysis paralysis, I say we go with your solution!

src/css/inputs.css
victorheringer: Fixed 5a51ae49c756a87276f6e8168b8d46471cf365cd

docs/inputs.mdx
victorheringer: Fixed 5a51ae49c756a87276f6e8168b8d46471cf365cd

docs/inputs.mdx
```diff
@@ -158,6 +159,31 @@ This input style does not have a disabled state.
...
+used to truncate the text size. Just put it like `<p class="a-text--secondary-small a-file-preview__item-text-truncate">` and it
+will use `text-overflow: ellipsis` to do the trick.
+
+<Playground>
+  <div className='a-file-preview'>
+    <div className='a-file-preview__item'>
+      <button aria-label='Remove'>
+        <i className='a-icon a-icon__x-men' />
+      </button>
+      <p class='a-text--secondary-small'>estratofinaceiro.csv</p>
```
lucasmazza: considering the docs are all in english, maybe this should be in english too?

src/css/inputs.css
```diff
@@ -114,6 +114,76 @@
...
+}
+
+.a-file-preview__item img {
+  height: 100%;
+  width: 100%;
+  object-fit: cover;
+  border-radius: 4px;
+}
+
+.a-file-preview__item img[src=''] {
```
lucasmazza: do we need this selector? shouldn't it be userland concern to not render src-less `img` tags? 🤔

docs/inputs.mdx
larimaza: Agreed, maybe "financialreport.csv"

docs/inputs.mdx
```diff
@@ -158,6 +159,31 @@ This input style does not have a disabled state.
   </div>
 </Playground>
 
+## uploaded files preview
+
+Used to display a preview for uploaded files.
+
+If you don't wanna handle the displayed text size by yourself like in the first example, we offer a `a-input--file__text-truncate` class that can be
+used to truncate the text size. Just put it like `<p class="a-text--secondary-small a-file-preview__item-text-truncate">` and it
+will use `text-overflow: ellipsis` to do the trick.
```
larimaza: This class needs to be renamed, so I went ahead and made a simpler rewrite suggestion:
```suggestion
If you need to truncate the displayed text, you may use the `a-file-preview__item--text-truncate` class in your `<p>` element.
```
I don't think we should explain in too many details in the docs because if we ever change the CSS, this will be outdated :)

src/css/inputs.css
```diff
@@ -114,6 +114,76 @@
...
+
+.a-file-preview__item p {
+  font-size: calc(var(--font-size-base) * 1.25);
+  padding: 12px;
+  margin: 0;
+  word-wrap: break-word;
+  max-width: 80px;
+}
+
+.a-file-preview__item p.a-file-preview__item-text-truncate {
```
larimaza: Looking at this closer now, maybe this class should be called "a-file-preview__item--text-truncate" with two dashes since it's a modifier.

src/css/inputs.css
victorheringer: This selector is not a must, but it serves at the same propose as this one:

```css
.a-file-preview__item p:empty {
  display: none;
}
``` 

It makes easier to use the same structure without toggling between `img` and `p` tags inside the structure with javascript (when generating it based on uploaded files), so this can direct represent some data structure like `const data = [{ text: '', image: ''}]` and we just need to plug the value.

```html
<div className='a-file-preview__item'>
  <button aria-label='Remove'>
    <i class='a-icon a-icon__x-men' />
  </button>
  <img src='' />
  <p></p>
</div>
```

This could be a little bit of over engineering 😅 but also could be handy, you think we should remove it and let developers handle the toggling thing  by themselves ?

lucasmazza: > This could be a little bit of over engineering 😅 but also could be handy, you think we should remove it and let developers handle the toggling thing by themselves ?

Yes 😬 in my opinion the design system shouldn't be doing a pass for developers emitting unnecessary/undesired markup.

docs/inputs.mdx
victorheringer: Fixed 8f3574acdc0f8ecdc9936a19463014874bc6b5a7

src/css/inputs.css
victorheringer: Fixed 8f3574acdc0f8ecdc9936a19463014874bc6b5a7

src/css/inputs.css
victorheringer: Agreed 😬, I removed both `p:empty` and `img[src='']`

Fixed 8ad9b3f0c4a561f55516e3729d34060b4d805cb4

PR #166, opened 2019-09-11
[BREAKING - DON'T MERGE UNTIL 2.0] Rename classes and adjust css to BEM conformity

larimaza:
# What

Review BEM conformity throughout the entire project (that will create massive breaking changes).

# Why

Some of our CSS is not following BEM standards and we're trying to fix this before the project grows even more.
We should also be more watchful of these patterns on future coding and code reviews.

# How

Rename a few classes;
Modify the Control Input button to create a standalone class for that very specific case, instead of having to use a combination of several classes.


larimaza:
[ch7511]

docs/js/inputs.js
```diff
@@ -130,64 +130,64 @@ const handleButtonDisable = buttonInputs => {
...
-            currentControlInput.value = '0,00';
-            return;
-          }
-
-          step = dataStep * -1;
-        }
-
-        currentControlInput.value = parseFloat(currentControlInputValue + step)
-          .toFixed(2)
+        );
```
victorheringer: This green marks are strange 🤔 , maybe this is tabs instead of spaces?

docs/inputs.mdx
```diff
@@ -181,11 +181,11 @@ To create the disabled state, add the `a-input--disabled` modifier class on elem
     <label id='controllabel1' htmlFor='control1'>
       Control input (R$)
     </label>
-    <button className='a-btn a-btn--ghost-uranus a-btn--small a-btn--icon a-input--control__decrement'>
-      <i className='a-icon a-icon__input-dash a-icon--size-small' />
+    <button className='a-input__btn-control'>
```
victorheringer: Just a question, if we're not using classes like `a-btn--ghost-uranus`, `a-input__btn-control` and etc,  we can change colors, size? Or changing color or size this way is not a desire scenario?

PR #170, opened 2019-09-13
Fix font CSS shorthands

elmarques:
# What

This PR removes an extra parameter in the font shorthands.

# Why

When we tested new Magnetis pages, we found that Microsoft Edge didn't display the fonts correctly.

Unlike other ~decent~ browsers, Edge doesn't handle/autocorrect extra parameters in the font shorthand.

Captura de Tela 2019-09-13 às 15 47 44

# How

Remove extra `normal` parameter from font shorthands.

# Sample

| Before | After |
| --- | --- |
| Captura de Tela 2019-09-13 às 16 55 53 | Captura de Tela 2019-09-13 às 16 54 19 |

# Refs

* [Clubhouse's card](https://app.clubhouse.io/magnetis/story/19038/corrigir-fontes-para-as-páginas-abertas-no-astro) [ch19038]
* [Font shorthand's ref](https://css-tricks.com/snippets/css/font-shorthand/)

### DIFF
--- a/src/css/fonts.css
+++ b/src/css/fonts.css
@@ -7,7 +7,7 @@
   --font-size-5base: calc(var(--font-size-base) * 5);
   --font-size-6base: calc(var(--font-size-base) * 6);
   --font-size-8base: calc(var(--font-size-base) * 8);
-  --font-primary: normal normal 600 normal var(--font-size-2base)/1.5 Poppins, sans-serif;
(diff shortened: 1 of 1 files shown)


elmarques:
@larimaza The problem is:

Captura de Tela 2019-09-16 às 16 47 57

🤣🤣🤣

Seriously now. I think that's exactly the problem. According to your reference, Edge doesn't support `font-strech` in font shorthand.

Thanks for sharing. I didn't know the exact reason.

PR #176, opened 2019-10-02
Add more z-index variables

brunobc182:
# What

This __PR__ adds more z-index variables.

# Why

Because we needed to use a higher z-index on case where we have various elements with deferents z-indexes.

### DIFF
--- a/src/css/z-index.css
+++ b/src/css/z-index.css
@@ -3,4 +3,10 @@
   --z-index-2: 20;
   --z-index-3: 30;
   --z-index-4: 40;
+  --z-index-5: 50;
+  --z-index-6: 60;
+  --z-index-7: 70;
+  --z-index-8: 80;
+  --z-index-9: 90;
+  --z-index-10: 100;
 }


elmarques:
:shipit: @AgtLucas 🦇@larimaza 

Hi guys! We don't really need all of them right now.

This use case came from the new home that demanded more than the 4 current levels.

Something like:

_menu on mobile_ ||||||| 6
_menu on desktop_ |||||| 5
_header content_ ||||| 4
_header background_ |||| 3
_footer on mobile_ |||| 3
_floating content_ ||| 2
_footer on desktop_ || 1
_rest of the world_ | none

We had a particular challenge overlaying the menu, content and footer on different screen sizes, and we use custom z-index values to deal with.

We thought it might be helpful for the future space fellows to leave some extra z-indexes, but if you think it's a non-lean approach, we can go back to the exact 6 we need.

What do you think, guys?

🎮🎮🎮🎮
```
press C to continue the merge with extra z-indexes
press D to delete the extra z-indexes
```

AgtLucas:
@elmarques Can I press both keys, like `CD`? 😂 

I really don't have any preference, whatever y'all decide, I'm onboard!

larimaza:
> :shipit: @AgtLucas 🦇@larimaza
> 
> Hi guys! We don't really need all of them right now.
> 
> This use case came from the new home that demanded more than the 4 current levels.
> 
> Something like:
> 
> _menu on mobile_ ||||||| 6
> _menu on desktop_ |||||| 5
> _header content_ ||||| 4
> _header background_ |||| 3
> _footer on mobile_ |||| 3
> _floating content_ ||| 2
> _footer on desktop_ || 1
> _rest of the world_ | none
> 
> We had a particular challenge overlaying the menu, content and footer on different screen sizes, and we use custom z-index values to deal with.
> 
> We thought it might be helpful for the future space fellows to leave some extra z-indexes, but if you think it's a non-lean approach, we can go back to the exact 6 we need.
> 
> What do you think, guys?
> 
> 🎮🎮🎮🎮
> 
> ```
> press C to continue the merge with extra z-indexes
> press D to delete the extra z-indexes
> ```

LOL @ the options, I'll press C 😅 
Thanks for the clarification, I was not aware of this complexity. I think we may leave the 10 z-indexes just to round them up, like you propose :)

PR #180, opened 2019-10-09
Chore/add stylelint order

LopesAlysson:
# What

Add [stylelint-order](https://github.com/hudochenkov/stylelint-order) plugin

# Why

We need to assure that css rules are written in a certain logical order.

# How

* add `stylelint-order` as a dev dependency on the project;
* update **src/*.css** files with `stylelint --fix`;


LopesAlysson:
[ch20528]

larimaza:
On a second note, maybe ordering the props by type and not alphabetically could make more sense when coding. For example, leaving all position-related settings close to each other.
I see that we need to declare the list how we'd like it ordered for Stylelint, so we could copy from GitHub's Primer:
https://github.com/primer/stylelint-config-primer/blob/master/property-order.js#L2-L170
(thanks @lucasmazza for the info)

LopesAlysson:
> On a second note, maybe ordering the props by type and not alphabetically could make more sense when coding. For example, leaving all position-related settings close to each other.
> I see that we need to declare the list how we'd like it ordered for Stylelint, so we could copy from GitHub's Primer:
> https://github.com/primer/stylelint-config-primer/blob/master/property-order.js#L2-L170
> (thanks @lucasmazza for the info)

Done in 6556630f3de95acb6ae90336b71b889f5d4b8bcb

LopesAlysson:
> The build is still complaining about formatting of the declaration order 🤔

This was about a change introduced on master that wasn't reflecting on my branch. Solved. :)

Issue #118, opened 2019-06-13
Links to codesandbox are broken


notnotgabriel:
# Description

Links to codesandbox direct me to an error page
`413. That’s an error.`
![image](https://user-images.githubusercontent.com/5247142/59442837-5f6a6400-8dd1-11e9-8a89-70685cdfa6b7.png)

# Expected behavior

Open codesandbox.io containing the component I selected.

# Steps to reproduce

Click in any codesandbox.io link.
![Screen Shot 2019-06-13 at 11 52 35](https://user-images.githubusercontent.com/5247142/59443131-d869bb80-8dd1-11e9-8ee3-3bfc73d6f2f0.png)

victorheringer:
As stated [here](https://github.com/magnetis/astro/pull/182), this is an docz problem and this functionally will be disabled by now!

PR #182, opened 2019-10-09
Remove code sanbox link

victorheringer:
# What

This __PR__ removes the link icon for code sanbox

# Why

Due to the get request for code sanbox api being to large, we have an error when we try to use this functionality. This is a docz problem where they should use a post request instead of get. We'll enable this in the future if any fix is published.

# How

* Change doczrc config

# Sample

Before

Captura de Tela 2019-10-09 às 16 25 15

After

Captura de Tela 2019-10-09 às 16 24 53

### DIFF
--- a/doczrc.js
+++ b/doczrc.js
@@ -27,5 +27,6 @@ export default {
     'iconography',
     'inputs',
     'typography'
-  ]
+  ],
+  codeSandbox: false
 };


victorheringer:
> Maybe we can close #118 then...

I think we can! Just ref this pr there and close it.

PR #183, opened 2019-10-09
Add manifesto illustrations

leandrothiers:
# What

Add manifesto illustration to improve our home presentation.

# Why

Why not?

# Sample

image


leandrothiers:
[ch20947]

docs/components/ManifestoCards.jsx
```diff
@@ -1,8 +1,14 @@
 import React from 'react';
+import atomic from '../img/atomic.png';
+import open from '../img/open.png';
+import accessible from '../img/accessible.png';
 
 export default () => (
   <div className='cards-box'>
     <div>
+      <div className="cards-box__image">
+        <img src={atomic} className="atomic-illustration" alt="atomic illustrative image" />
```
AgtLucas: There's no need to use the word `image` inside an `alt` attribute, 'cause screen reader would read this as `Image: atomic illustrative image` so it's repeating, which is not ideal.

Also, I think we could pass an empty string for these alts here, these images are just for presentation purpose, so there's no need to add more text for the user.

larimaza:
I see @LopesAlysson is teaching you his secret ways...
("Why not?" 🤣)

docs/components/ManifestoCards.jsx
larimaza: True, sometimes with good intentions we can over-inform our screen reader user 😬

leandrothiers: Done!

AgtLucas: @leandro Sorry, I meant you still need to pass the `alt` attribute, like so:

```html
<img src='foo.png' alt="" />
```

:)

leandrothiers: My bad. Sorry for distraction.

leandro: @AgtLucas did you call me, sir? 🙈😅

AgtLucas: @leandro Oh damn! Sorry, wrong Leandro!

Still, y'all are 5 Leandros behind 😂 

There are 7 Lucas here and 2 Leandros :P

leandro: @AgtLucas Don't worry. I'm used to this. I even get mentioned in random projects on github sometimes.

PR #186, opened 2019-10-10
Add A11y eslint plugin

LopesAlysson:
# What

Add A11y plugin to make sure we follow the recommended approach on accessibility;

# Why

It's important to make our design system accessible to everyone.

# How

* add `eslint`, `babel-eslint` and `eslint-plugin-jsx-a11y`;
* create `.eslintrc` file;
* run eslint and fix errors;

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,11 @@
+{
+  "plugins": ["jsx-a11y"],
+  "extends": ["plugin:jsx-a11y/strict"],
+  "parser": "babel-eslint",
+  "parserOptions": {
+    "ecmaFeatures": {
+      "jsx": true,
+      "modules": true
+    }
+  }
+}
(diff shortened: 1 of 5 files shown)


LopesAlysson:
[ch20990]

.eslintrc
```diff
@@ -0,0 +1,11 @@
+{
+  "plugins": ["jsx-a11y"],
+  "extends": ["plugin:jsx-a11y/strict"],
```
AgtLucas: Humm, I wonder if make sense to add such plugin here, as Astro doesn't have any JSX itself, only the docz page...

🤔

AgtLucas: Maybe it's better to move this to `docz` directory.

LopesAlysson: As discussed in our slack channel, we will stick to the root file.

PR #191, opened 2019-10-23
Adjust spacing in home screen

leandrothiers:
# What

The space below "get started" title in home screen got too big after inserting the illustrations in home screen. In this PR it was fixed.

# Why

To improve astro home page design

# How

The problem was css classes conflicting and the solution was create a specific rule to attend the illustration spacing.

# Sample

image

### DIFF
--- a/docs/components/ManifestoCards.jsx
+++ b/docs/components/ManifestoCards.jsx
@@ -4,7 +4,7 @@ import open from '../img/open.png';
 import accessible from '../img/accessible.png';
 
 export default () => (
-  <div className='cards-box'>
+  <div className='cards-box cards-box--spaced'>
     <div>
       <div className="cards-box__image">
         <img src={atomic} className="atomic-illustration" alt="" />
--- a/docs/css/home.css
+++ b/docs/css/home.css
@@ -4,6 +4,9 @@
   display: flex;
   flex-wrap: wrap;
   justify-content: space-between;
+}
+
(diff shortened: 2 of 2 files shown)


docs/css/home.css
```diff
@@ -4,6 +4,9 @@
   display: flex;
   flex-wrap: wrap;
   justify-content: space-between;
+}
+
+.cards-box.spacing {
```
LopesAlysson: Maybe we could use the BEM convention and name it something like a `spaced` modifier.
```suggestion
.cards-box--spaced {
```
What do you think?

docs/components/ManifestoCards.jsx
```diff
@@ -4,7 +4,7 @@ import open from '../img/open.png';
 import accessible from '../img/accessible.png';
 
 export default () => (
-  <div className='cards-box'>
+  <div className='cards-box spacing'>
```
larimaza: May I suggest the class name "cards-box--spacing"? I think it makes more sense with our BEM guidelines :) And then you don't have to combine them in the CSS sheet

leandrothiers: 909e8bf

docs/css/home.css
leandrothiers: 909e8bf

larimaza:
[ch21701]

PR #194, opened 2019-10-28
Remove all animations and transitions for people that don't wanna see them

AgtLucas:
# What

Have you ever felt nauseous or dizzy while navigating through a website or using an app with too much aggressive animation, or parallax effects? Well, maybe you suffer from a vestibular disorder…

I wrote a blog post about this last year: https://lcs.sh/blog/2018/01/28/reduce-motion/

**THIS ONLY REMOVE FROM PEOPLE THAT ENABLE THE REDUCTION ON THEIR DEVICES.**

# Why

To improve Astro's a11y. :)

# How

Using `@media (prefers-reduced-motion: reduce)`.

### DIFF
--- a/src/css/base.css
+++ b/src/css/base.css
@@ -26,3 +26,14 @@ ul {
   padding: 0;
   list-style: none;
 }
+
+/* Remove all animations and transitions for people that don't wanna see them */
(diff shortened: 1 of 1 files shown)


larimaza:
Just... WOW!

PR #198, opened 2019-11-08
Table styles

lucasmazza:
# What

This brings the table styles from recent app work into astro to pave the way as a set of defaults for any purpose tables 

## Things to do (please pitch in!)

* [x] Review spacing rules to keep consistency with existing components, as all the `padding-` settings came from one single design, I don't know if we have strict rules for this.
* [x] Review the purpose of `a-table__row-actions`
* [x] Write an actual documentation to go along with the `<Playground />` example
* ~~Should we add any responsive styles right now? [There are a handful of strategies for this problem](https://css-tricks.com/responsive-data-table-roundup/), but I don't know if we should have a default approach right now~~

# Sample

Screen Shot 2019-11-08 at 16 22 05

### DIFF
--- a/docs/tables.mdx
+++ b/docs/tables.mdx
@@ -1,0 +1,106 @@
+---
+name: tables
+---
+
+import '../src/index.css';
+import { Playground } from 'docz';
+
+# tables
+
+### standard table styles
+
+To apply the standard styles to a `table` element, add the `a-table` class. No
+modifiers are available or required.
+
+Be aware that typgraphy styles are not applied by default and you should style
+the `th` and `td` elements as necessary for your table.
+
+Buttons using the `a-btn` (and related modifiers) that are placed inside the last
+`td` of each row will be properly aligned together as the set of actions available
+for that row.
+
+<Playground>
+  <table className='a-table'>
+    <thead>
+      <tr>
+        <th className='a-text--medium' scope='col'>rank</th>
+        <th className='a-text--medium' scope='col'>title</th>
+        <th className='a-text--medium' scope='col'>distributor</th>
+        <th className='a-text--medium' scope='col' colspan='2'>worldwide gross</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr>
+        <td className='a-text--secondary-medium'><strong>1</strong></td>
+        <td className='a-text--secondary-medium'>Avengers: Infinity War</td>
+        <td className='a-text--secondary-medium'>Disney</td>
+        <td className='a-text--secondary-medium'>$2,048,359,754</td>
+        <td>
+          <button title='delete' aria-label='delete' className='a-btn a-btn--ghost-mars a-btn--medium a-btn--icon'>
+            <i className='a-icon a-icon--trash a-icon--size-medium' />
+          </button>
+          <a  title='edit' aria-label='edit' className='a-btn a-btn--ghost-uranus a-btn--medium a-btn--icon'>
+            <i className='a-icon a-icon--pencil a-icon--size-medium' />
+          </a>
+          <a  title='view' aria-label='view' className='a-btn a-btn--ghost-uranus a-btn--medium a-btn--icon'>
+            <i className='a-icon a-icon--circle-right a-icon--size-medium' />
+          </a>
+        </td>
+      </tr>
+
+      <tr>
+        <td className='a-text--secondary-medium'><strong>2</strong></td>
+        <td className='a-text--secondary-medium'>Black Panther</td>
+        <td className='a-text--secondary-medium'>Disney</td>
+        <td className='a-text--secondary-medium'>$1,346,913,161</td>
+        <td>
+          <button title='delete' aria-label='delete' className='a-btn a-btn--ghost-mars a-btn--medium a-btn--icon'>
+            <i className='a-icon a-icon--trash a-icon--size-medium' />
+          </button>
+          <a  title='edit' aria-label='edit' className='a-btn a-btn--ghost-uranus a-btn--medium a-btn--icon'>
+            <i className='a-icon a-icon--pencil a-icon--size-medium' />
+          </a>
+          <a  title='view' aria-label='view' className='a-btn a-btn--ghost-uranus a-btn--medium a-btn--icon'>
+            <i className='a-icon a-icon--circle-right a-icon--size-medium' />
+          </a>
+        </td>
+      </tr>
+
+      <tr>
+        <td className='a-text--secondary-medium'><strong>3</strong></td>
+        <td className='a-text--secondary-medium'>Jurassic World: Fallen Kingdom</td>
+        <td className='a-text--secondary-medium'>Universal</td>
+        <td className='a-text--secondary-medium'>$1,309,484,461</td>
+        <td>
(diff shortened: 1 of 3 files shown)


src/css/tables.css
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.a-table tr:nth-child(odd) td {
+  background-color: var(--color-moon-100);
+}
+
+.a-table td.a-table__row-actions {
+  padding-right: 24px;
+  text-align: right;
+  display: flex;
```
lucasmazza: this flex styles care around when I tried multiple links/buttons together - it can be extracted to some `button-group` kind of component and removed from the table styles

src/css/tables.css
```diff
@@ -0,0 +1,48 @@
...
+.a-table th:last-child {
+  padding-right: 64px;
+}
+
+.a-table tr:nth-child(odd) td {
+  background-color: var(--color-moon-100);
+}
+
+.a-table td.a-table__row-actions {
+  padding-right: 24px;
```
lucasmazza: I have some mixed feelings of having uneven paddings on the row, might be worth to review this with our designers to see if we keep this around

src/css/tables.css
```diff
@@ -0,0 +1,48 @@
+.a-table {
+  width: 100%;
+  border-collapse: collapse;
+}
+
+.a-table th {
+  padding-top: 16px;
```
lucasmazza: should we infer theses spacing rules from `--font-size-base` & friends?  Everything is on a `8px` scale

larimaza:
Nice job, thanks for contributing ✨ 
About the to-dos in the description:

- I would get definitive spacing rules and responsive behaviour from the design team, as they are the ones responsible for creating visual patterns. And yes, I would add the responsive behaviour here, by asking them to provide a default. Once we document it, they'll stick to that default in the future;

- Actual documentation to go in the website is currently a combination of the design copy that they provide + our technical instructions. We can start with only usage instructions and add their copy in the future.

src/css/tables.css
larimaza: We could, but I think we should create a technical debt to implement these variables in the whole project, instead of doing it just here 🤔

src/css/tables.css
larimaza: I'm all for questioning the specs, +1

src/css/tables.css
larimaza: While we don't have any predicted use for a button-group component, do you think we can leave them here for now, and extract them once we need them in another context?

docs/tables.mdx
```diff
@@ -0,0 +1,94 @@
...
+            <i className='a-icon a-icon--pencil a-icon--size-medium' />
+          </a>
+          <a  title='view' className='a-btn a-btn--ghost-uranus a-btn--medium a-btn--icon'>
+            <i className='a-icon a-icon--circle-right a-icon--size-medium' />
+          </a>
+        </td>
+      </tr>
+
+      <tr>
+        <td className='a-text--secondary-medium'><strong>3</strong></td>
```
larimaza: Did you use `<strong>` for semantic purposes or only for bold weight?
I ask because I've been thinking of creating a font modifier class to add the bold weight to Lato, it's not official but it's very commonly used

lucasmazza: more semantics than styling

docs/tables.mdx
```diff
@@ -0,0 +1,94 @@
...
+        <td className='a-table__row-actions'>
+          <button title='delete' className='a-btn a-btn--ghost-mars a-btn--medium a-btn--icon'>
+            <i className='a-icon a-icon--trash a-icon--size-medium' />
+          </button>
+          <a  title='edit' className='a-btn a-btn--ghost-uranus a-btn--medium a-btn--icon'>
+            <i className='a-icon a-icon--pencil a-icon--size-medium' />
+          </a>
+          <a  title='view' className='a-btn a-btn--ghost-uranus a-btn--medium a-btn--icon'>
+            <i className='a-icon a-icon--circle-right a-icon--size-medium' />
+          </a>
```
AgtLucas: While I understand that this is not in the PR's scope, we could add an `aria-label` alongside these `title` attribute.

docs/tables.mdx
```diff
@@ -0,0 +1,94 @@
...
```
AgtLucas: Same here.

lucasmazza:
we did a quick review with the design team to polish some of the alignment bits and discussed that we won't be tackling responsiveness right away as the current use case does not require it. I've also added some quick instructions to the `mdx` file to go along with the playground snippet

docs/tables.mdx
```diff
@@ -0,0 +1,106 @@
...
+        </td>
+      </tr>
+
+      <tr>
+        <td className='a-text--secondary-medium'><strong>2</strong></td>
+        <td className='a-text--secondary-medium'>Black Panther</td>
+        <td className='a-text--secondary-medium'>Disney</td>
+        <td className='a-text--secondary-medium'>$1,346,913,161</td>
+        <td>
+          <button title='delete' aria-title='delete' className='a-btn a-btn--ghost-mars a-btn--medium a-btn--icon'>
```
AgtLucas: @lucasmazza It should be `aria-label` instead of `aria-title`. :sorriso-triste:

docs/tables.mdx
```diff
@@ -0,0 +1,106 @@
...
+import { Playground } from 'docz';
+
+# tables
+
+### standard table styles
+
+To apply the standard styles to a `table` element, add the `a-table` class. No
+modifiers are available or required.
+
+Be aware that typgraphy styles are not applied by default and you should style
```
larimaza:
```suggestion
Be aware that typography styles are not applied by default and you should style
```

docs/tables.mdx
```diff
@@ -0,0 +1,106 @@
...
+the `th` and `td` elements as necessary for your table.
+
+Buttons using the `a-btn` (and related modifiers) that are placed inside the last
```
larimaza: Love this!

docs/tables.mdx
```diff
@@ -0,0 +1,106 @@
...
+`td` of each row will be properly aligned together as the set of actions available
+for that row.
```
larimaza: Love this!

src/css/tables.css
leandro: Why don't we use `rem` for this kind of measures? Normally (by default) `1rem` equals `16px`. It's a good starting point.

larimaza: `rem` is relative to the :root, right? Doesn't that mean sizes would vary according to the project Astro is being used on?

lucasmazza: @larimaza yes - I think we would need to have a fixed `font-size` on `body`/ a root selector to make `rem` usage predictable.

AgtLucas: I like `rem`. You should discuss this, maybe we could start from scratch with `rem` on Astro Galaxy.

leandro: @larimaza Yes. Sorry for the delayed response. In mostly all browsers (including mobiles', afaik), the default root/body's `font-size` is `16px`, which already would be suitable for us in terms of being multiple of `8px`, as long as we use multiples of `0.5rem` in our margins and paddings.

PR #218, opened 2020-01-08
Adjust checkbox positioning

brunojabs:
# What

Adjusts the positioning of checkbox

# Why

There's a bug when the label is larger than 1 line, also the checkbox is not perfect centered relative to the label

Before | After
------------ | -------------
Astro | Astro

### DIFF
--- a/src/css/checkbox.css
+++ b/src/css/checkbox.css
@@ -3,7 +3,6 @@
 }
 
 .a-checkbox {
-  position: relative;
   display: flex;
   align-items: center;
 }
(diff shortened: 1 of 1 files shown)


src/css/checkbox.css
```diff
@@ -25,12 +24,16 @@
...
-  margin: 4px 0;
   cursor: pointer;
   opacity: 0;
   order: 1;
 }
 
+.a-checkbox__shape {
+  position: relative;
+  line-height: 0;
+}
```
AgtLucas: Do we really need this?

brunojabs: Maybe. 

Without the relative, the absolute for the `::after` does not work properly
Without setting the line-height to the checkbox stay on the top of the line

I don't know if there's a better solution, I accept halp

